PR #36, opened 2014-05-31
Default normed quaternions

### LINKED ISSUE #35: Default Normalization of Quaternion
This is probably a fellow up to
https://github.com/flexible-collision-library/fcl/pull/26

I would like to remove the possibility of assigning non-normalized quaternions, since they are not used in this context anyway, and i am running into rounding errors when running long enough (i.e. quaternions become unnormalized after time and crash my program).

I therefore removed the explicit initiation of a non-normed quaternion, which i think was a crude hack anyway
https://github.com/flexible-collision-library/fcl/pull/34

So now i would like to make them fail-safe, i.e. that quaternions are normed by default. In this way, future bugs concerning the norm would be prevented.

aorthey:
Fixed #35

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/36)

### DIFF
--- a/src/math/transform.cpp
+++ b/src/math/transform.cpp
@@ -58,9 +58,9 @@ Quaternion3f& Quaternion3f::normalize()
       data[2] = 0;
       data[3] = 0;
     } else {
-      double sl = sqrtf(l);
+      double sl = 1.0/std::sqrt(l);
       for (uint i = 0 ; i < 4 ; i++) {
-        data[i] /= sl;
+        data[i] *= sl;
       }
     }
   }


src/math/transform.cpp
```diff
@@ -41,6 +41,32 @@
...
+      data[i] /= 0.5*(1+l);
+    }
+  } else {
+    if ( l < 1e-6 ) {
+      data[0] = 1;
+      data[1] = 0;
+      data[2] = 0;
+      data[3] = 0;
+    } else {
+      double sl = sqrtf(l);
```
sherm1: This should be `std::sqrt(l)` rather than `sqrtf` which assumes single precision floats.

src/math/transform.cpp
```diff
@@ -41,6 +41,32 @@
...
+      for (uint i = 0 ; i < 4 ; i++) {
+        data[i] /= sl;
```
sherm1: Division is *much* slower than multiplication. So for speed, it would be better to precalculate one over sqrt(l) and multiply here.

jslee02:
Is this ready to merge?

sherm1:
Sorry to be a drag, but I don't think so. This is tricky code with hard-coded constants and I don't see any unit tests for it. The phrase that comes to mind is "premature optimization". Is this really faster than a simple 1/sqrt normalization? Does it work properly for float and double precision (assuming it should)? If there really is a significant performance gain (measured!) then there should be careful tests of all the various cases to make sure they work and that the constants have been chosen correctly. There should also be comments explaining why these are the right choices.

Absent that additional work, I would much rather see a simple method that gets the job done with fewer risks.

---

Reviewed 1 of 2 files at r4.
Review status: 1 of 2 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[src/math/transform.cpp, line 51 at r4](https://beta.reviewable.io/reviews/flexible-collision-library/fcl/36#-L8ZTGjDB_ibBwJHJB_G-r4-51:-L9SZTrLAnQxDqHI37em:brbtlxt) ([raw file](https://github.com/flexible-collision-library/fcl/blob/fca881eb5f1aa3416c940fbbcf631e1c9c5583e7/src/math/transform.cpp#L51)):*
<details><summary><i>Previously, isucan (Ioan A Sucan) wrote…</i></summary><blockquote>

I like this epsilon optimization, I did not know about it!
</blockquote></details>

Wait ... have you measured the performance? To me this looks like it would be slower than a single square root. This one does 4 divides, which are _much_ more expensive than a floating point add or multiply (depending on CPU, could be 6-10X). A sqrt is likely to be 20-25X so four divides could be very close. Much simpler would be to calculate `1/sqrt(l)` and then multiply four times -- are you sure this complicated approach is faster than that? 

Another question is whether this is valid for `float` precision, which I believe is an FCL option (please correct me if I'm wrong). If Quaternion3f were float rather than double, the case guarded by double-precision epsilon would never get executed. 

Something this tricky should be tested carefully before adoption. 

---

*[src/math/transform.cpp, line 53 at r4](https://beta.reviewable.io/reviews/flexible-collision-library/fcl/36#-L8ZTGjDB_ibBwJHJB_G-r4-53:-L9S_xNmDT2suBnx0MFz:b-8f427e) ([raw file](https://github.com/flexible-collision-library/fcl/blob/fca881eb5f1aa3416c940fbbcf631e1c9c5583e7/src/math/transform.cpp#L53)):*
<details><summary><i>Previously, orthez wrote…</i></summary><blockquote>

Implementation uses an approximation which avoids the sqrt execution and is therefore much more efficient. More infos are here: http://stackoverflow.com/a/12934750. 
</blockquote></details>

As I read the stackoverflow, the author there does only a single divide, not 4!

---

*[src/math/transform.cpp, line 55 at r4](https://beta.reviewable.io/reviews/flexible-collision-library/fcl/36#-L9SchyK87y8nwS3ICWT:-L9SchyK87y8nwS3ICWU:bdb40qu) ([raw file](https://github.com/flexible-collision-library/fcl/blob/fca881eb5f1aa3416c940fbbcf631e1c9c5583e7/src/math/transform.cpp#L55)):*
> ```cpp
>     }
>   } else {
>     if ( l < 1e-6 ) {
> ```

It doesn't seem appropriate to have a hard-coded tolerance here if the precision can vary (float or double). This case appears to be assuming float precision but even then it seems odd since "l" is squared precision -- that means if the norm is 1e-3 or smaller this returns an identity quaternion. Is that the right thing to do? Certainly needs a comment explaining why, if so.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/flexible-collision-library/fcl/36#-:-L9SdXJy2RQ1sLWnaJ4N:b-9id10k)*

PR #42, opened 2014-12-01
Fixed BoxBox Intersection Algorithm

EdsterG:
The boxBox intersection algorithm had bugs in it. After comparing it to the bullet implementation, everything is now fixed. When there are multiple points of collision between two boxes, the deepest point of collision is returned.

### DIFF
--- a/src/narrowphase/narrowphase.cpp
+++ b/src/narrowphase/narrowphase.cpp
@@ -1038,7 +1038,7 @@ int boxBox2(const Vec3f& side1, const Matrix3f& R1, const Vec3f& T1,
   // separating axis = u1 x (v1,v2,v3)
   tmp = pp[2] * R(1, 0) - pp[1] * R(2, 0);
   s2 = std::abs(tmp) - (A[1] * Q(2, 0) + A[2] * Q(1, 0) + B[1] * Q(0, 2) + B[2] * Q(0, 1));
-  if(s2 > eps) { *return_code = 0; return 0; }
+  if(s2 > 0) { *return_code = 0; return 0; }
   n = Vec3f(0, -R(2, 0), R(1, 0));
   l = n.length();
   if(l > eps) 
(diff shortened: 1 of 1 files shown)


hsu:
just for my own reference, a unit test might go [here?](https://github.com/flexible-collision-library/fcl/blob/master/test/test_fcl_geometric_shapes.cpp#L202)

scpeters:
@mikeferguson I can assist with writing unit tests if you can give some hints about how you did the testing and debugging for these changes. I would most prefer to write a test that fails on the master branch but passes with this branch.

jslee02:
I wrote a simple test that fails on the master branch but passes with this branch.
```cpp
void testBoxBoxPullRequest42(const Matrix3f& R)
{
  Box s1(100, 100, 100);
  Box s2(10, 20, 30);

  // Vertices of s2
  std::vector<Vec3f> vertices(8);
  vertices[0].setValue( 1,  1,  1);
  vertices[1].setValue( 1,  1, -1);
  vertices[2].setValue( 1, -1,  1);
  vertices[3].setValue( 1, -1, -1);
  vertices[4].setValue(-1,  1,  1);
  vertices[5].setValue(-1,  1, -1);
  vertices[6].setValue(-1, -1,  1);
  vertices[7].setValue(-1, -1, -1);

  for (int i = 0; i < 8; ++i)
  {
    vertices[i][0] *= 0.5 * s2.side[0];
    vertices[i][1] *= 0.5 * s2.side[1];
    vertices[i][2] *= 0.5 * s2.side[2];
  }

  Transform3f tf1 = Transform3f(Vec3f(0, 0, -50));
  Transform3f tf2 = Transform3f(R);

  Vec3f normal;
  Vec3f point;
  double penetration;

  // Make sure contact
  bool res = solver1.shapeIntersect(s1, tf1, s2, tf2, &point, &penetration, &normal);
  BOOST_CHECK(res);

  // Compute global vertices
  for (int i = 0; i < 8; ++i)
    vertices[i] = tf2.transform(vertices[i]);

  // Sort the vertices so that the lowest vertex along z-axis comes first
  std::sort(vertices.begin(), vertices.end(), compareContactPoints);

  // The lowest vertex along z-axis should be the contact point
  BOOST_CHECK(vertices[0].equal(point));
}

BOOST_AUTO_TEST_CASE(pull_request_42)
{
  FCL_UINT32 numTests = 1e+3;
  Transform3f tf;

  for (FCL_UINT32 i = 0; i < numTests; ++i)
  {
    generateRandomTransform(extents, tf);
    testBoxBoxPullRequest42(tf.getRotation());
  }
}
```

To perform fair tests with master branch, bool boxBoxIntersect() need some modification which was made in this pull request. The master branch's `boxBoxIntersect()` returns averaged contact point while this PR's version returns deepest contact point. 

`boxBox2()` function cannot be called outside of narrowphase.cpp (the header doesn't have declaration) it wasn't possible test boxBox2() individually.

Issue #25, opened 2014-02-17
broken unit tests


mamoll:
in 0.3.0 the following two unit tests are broken: test_fcl_broadphase and test_fcl_shape_mesh_consistency. In the first one there might some sort of memory corruption.

jslee02:
**test_fcl_broadphase** and **test_fcl_shape_mesh_consistency** seems to work now but [other 4 tests are failing](https://travis-ci.org/dartsim/fcl/jobs/53141229#L882). As a side note, **test_fcl_simple** doesn't work on only debug mode.

jslee02:
The current master branch passes all the unit tests so I believe this issue is addressed. Please reopen if it's not.

PR #46, opened 2015-03-05
Support Travis CI testing

jslee02:
This pull request enables to perform build tests for FCL on [Travis CI](https://travis-ci.org). To make this work, GitHub Webhook should be activated by one of the maintainers of this repository. Here is a [short instruction](http://docs.travis-ci.com/user/getting-started/#Step-two%3A-Activate-GitHub-Webhook).

Some tests take too much time on debug mode for Travis-CI. Travis-CI regards a test is failed if there is no response for more than 10 minutes. I reduced the size of the tests not to take more than 10 minutes.

There were some broken tests (see #25), and they are disabled for now. These broken tests will be tracked in #25.


jslee02:
To enable to test on Mac, [we need to send an email to Travis-CI](http://blog.travis-ci.com/2014-05-13-multi-os-feature-available/). I think GitHub Webhook should be activated first for this.

mikeferguson:
As I mentioned in the other ticket -- can we clean up the history a bit? There is a merge of upstream as well as commit/reverts in this history. This should really be like 2 commits: 1 that adds the travis-CI, and a second commit which reduces the sizes of tests and disables the borked tests.

jslee02:
Replace by #49.

PR #47, opened 2015-03-05
Support multiple contacts for primitive shape collision

### LINKED ISSUE #15: Multiple contact points for primitive shapes
I'm Jeongseok Lee a member of DART development team. DART is powered by FCL for rigid body collision detection. We are trying to fully utilize FCL.

DART is using meshes for primitive shapes and now we want to use FCL's primitive shapes to reduce the computational cost and get analytical contact results. We however have an issue that FCL returns only one contact point for primitive shape collision which is a critical to us. 

In some cases, multiple contact points are necessary in dynamics simulation. For example, when we lay a box on the ground (let's assume that the ground is thin box shape), we need at least three contact points to support the box. With one contact point at the center of the box's bottom face, it is unstable and can be fall down through the ground.

FCL gets contact point from GJKSolver_indep::shapeIntersect() or GJKSolver_libccd::shapeIntersect(). The shapeIntersect() function calls boxBoxIntersect() and boxBoxIntersect() gets multiple contact points from boxBox2() but returns the average of the contact points. (see: narrowphase.cpp ln:1350)

~~~cpp
bool boxBoxIntersect(const Box& s1, const Transform3f& tf1,
                     const Box& s2, const Transform3f& tf2,
                     Vec3f* contact_points, FCL_REAL* penetration_depth_, Vec3f* normal_)
{
  std::vector<ContactPoint> contacts;
  int return_code; 
  Vec3f normal;
  FCL_REAL depth;
  /* int cnum = */ boxBox2(s1.side, tf1.getRotation(), tf1.getTranslation(),
                           s2.side, tf2.getRotation(), tf2.getTranslation(),
                           normal, &depth, &return_code,
                           4, contacts);

  if(normal_) *normal_ = normal;
  if(penetration_depth_) *penetration_depth_ = depth;

  if(contact_points)
  {
    Vec3f contact_point;
    for(size_t i = 0; i < contacts.size(); ++i)
    {
      contact_point += contacts[i].point;
    }

    contact_point = contact_point / (FCL_REAL)contacts.size();

    *contact_points = contact_point;
  }

  return return_code != 0;
}
~~~

It may not easy to decide which set of contact points is correct depend on colliding primitive shape combinations. However, it would be nice if FCL returns multiple contact points and the user enable to select contact points what the user wants. 

Please let me know if I'm misunderstanding about FCL, thanks!

jslee02:
This pull request enables FCL to return multiple contacts per primitive-primitive collision. This feature is essential when FCL is used in physics simulation. For example, if we want to stack a primitive shape on another primitive shape in presence of gravity, multiple contacts are required to support the upper object. Single contact will make it unstable.

As I known, the most low level collision object types in FCL are triangle and primitive shapes. Depending on the kind of primitive shape, FCL uses different collision algorithms. Here is the category of collision algorithms in FCL: 
1. triangle-triangle algorithm
2. primitiveShape-triangle algorithms
3. general (convex) primitiveShape-primitiveShape algorithm using libccd
4. bunch of fast algorithms for specific primitiveShape-primitiveShape combination

All the collision algorithms returns single contact point. If a mesh and other collision object are colliding, multiple contacts can be detected because a mesh consists of many of triangles. The problem is the case that two primitive shapes are colliding. Although the algorithm for specific primitive shapes combination (e.g, box-box), it had to pick a single contact or average them because of the limitation of API. So I propose to change the API to be able to return multiple contacts as:
```cpp
// Previous API
template<typename S1, typename S2>
bool shapeIntersect(const S1& s1, const Transform3f& tf1, const S2& s2, const Transform3f& tf2, Vec3f* contact_points, FCL_REAL* penetration_depth, Vec3f* normal*) const FCL_DEPRECATED;

// Proposing API
template<typename S1, typename S2>
bool shapeIntersect(const S1& s1, const Transform3f& tf1, const S2& s2, const Transform3f& tf2, std::vector<ContactPoint>*) const;
```

As of now, box-box collision algorithm is the only one can detect multiple contacts. Other collision algorithm for different combination of primitive shapes needs to be implemented. This will be done in following pull requests.

Since the previous API, which is exposed to the user, is just marked as deprecated instead of removed, this PR doesn't hurt the backward compatibility. With this new API, [all the unit tests (those were not broken as itself) are passing](https://travis-ci.org/dartsim/fcl).

This pull request also fixes #15.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,41 @@
+language: cpp
+
+os:
+  - linux
+  - osx
+
+compiler:
+  - gcc
+  - clang
+
+env:
+  - BUILD_TYPE=Debug
+  - BUILD_TYPE=Release
+
+matrix:
+  exclude:
+    - os: osx
+      compiler: gcc
+
+install:
+  # Install dependencies for FCL
+  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then 'ci/install_linux.sh' ; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx"   ]; then 'ci/install_osx.sh'   ; fi
+
+script:
+  # Create build directory
+  - mkdir build
+  - cd build 
+
+  # Configure
+  - cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE ..
+
+  # Build
+  - make -j4
+
+  # Run unit tests
+  - make test
+
+  # Make sure we can install with no issues
+  - sudo make -j4 install
+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,19 @@ if (NOT CMAKE_BUILD_TYPE)
    set(CMAKE_BUILD_TYPE Release)
 endif()
 
+# Set build type variable
+set(BUILD_TYPE_RELEASE FALSE)
+set(BUILD_TYPE_DEBUG FALSE)
+
+string(TOUPPER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE_UPPERCASE)
+if("${CMAKE_BUILD_TYPE_UPPERCASE}" STREQUAL "RELEASE")
+  set(BUILD_TYPE_RELEASE TRUE)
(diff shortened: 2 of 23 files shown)


jslee02:
Replace by #50

PR #48, opened 2015-03-06
Support ellipsoid shape

### LINKED ISSUE #45: Add ellipsoid support in FCL

jslee02:
This pull request adds ellipsoid shape to FCL.

New unit tests were added and seems [Travis CI is happy with them](https://travis-ci.org/dartsim/fcl/builds/53305138).

This fixes #45 as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,41 @@
+language: cpp
+
+os:
+  - linux
+  - osx
+
+compiler:
+  - gcc
+  - clang
+
+env:
+  - BUILD_TYPE=Debug
+  - BUILD_TYPE=Release
+
+matrix:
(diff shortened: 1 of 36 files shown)


mikeferguson:
There are appear to be a lot of commits in common between this and #47 and #46. There are also merges of upstream. Can we maybe clean up the history of these PRs? (ideally each would be only one or a few commits, and each PR would be independent of one another (we can of course work through each one in order and then you can rebase away once each gets into the mainline)). 35 commits with 3000 additions is really challenging to review.

jslee02:
I see. Let me recreate new three pull requests, which are independent each other.

jslee02:
Replaced by #51

PR #51, opened 2015-03-06
Support ellipsoid shape

jslee02:
This pull request adds ellipsoid shape support.

Note: This pull request already includes the changes of #52.

PR #53, opened 2015-03-07
Use same shape algorithms for the reverse shape pair

jslee02:
FCL uses two kinds of shape algorithms. One is Minkowski Portal Refinement (MPR) algorithm of libccd (or built-in GJK implementation) for general convex shapes. The other one is a set of special algorithms for some shape pairs. In general, the special algorithms are faster than MPR algorithm so if FCL has special algorithm for certain shape pair than use it instead of MPR algorithm.

The special algorithm functions have an order to take shape pair. For example, `capsulePlaneIntersect` first takes `Capsule` and then `Plane`. The problem I found is that FCL dose not use the special algorithm for the reverse shape pair. That means FCL calls `capsulePlaneIntersect` for capsule-plane pair but calls MPR function for plane-capsule pair. Collision detection works well even though FCL uses different algorithm for the reverse shape pair but it shows slightly different results. But it seems like unnecessary differences to me.

This PR makes FCL to use same shape algorithms for the reverse shape pair as well.

This PR also includes unit tests that checks whether the special shape algorithms are called for the reverse case as well.


test/test_fcl_geometric_shapes.cpp
```diff
@@ -2752,3 +2752,177 @@ BOOST_AUTO_TEST_CASE(conecone)
...
+  Cylinder cylinder(5, 10);
+  Plane plane(Vec3f(), 0.0);
+  Halfspace halfspace(Vec3f(), 0.0);
+
+  // Use sufficiently short distance so that all the primitive shapes can intersect
+  FCL_REAL distance = 5.0;
+
+  // If new shape intersection algorithm is added for two distinct primitive
+  // shapes, uncommnet associated lines. For example, box-sphere intersection
+  // algorithm is added, then uncommnet box-sphere.
```
scpeters: `uncommnet` -> `uncomment` here and the line above. Maybe do a search for any other instances of the misspelling

src/narrowphase/narrowphase.cpp
```diff
@@ -2555,12 +2555,46 @@ bool planeIntersect(const Plane& s1, const Transform3f& tf1,
 
 } // details
 
+// Shape intersect algorithms not using libccd
+//
+// +------------+-----+-----------+--------+---------+------+----------+-------+------------+----------+
+// |            | box | ellipsoid | sphere | capsule | cone | cylinder | plane | half-space | triangle |
```
scpeters: ellipsoids aren't in the master branch yet, though they are in #51 

I'm ok with them having a slot in this table though, to save on editing the ascii art

jslee02: Removed ellipsoid for master branch. I think it's doable to add back.

test/test_fcl_geometric_shapes.cpp
jslee02: Fixed them in d277922. There were 4 the misspelling total.

scpeters:
These seem like reasonable changes; the main challenge for me as a reviewer is the size of the diff. After looking at it a few more times, I should be ready to approve.

jslee02:
The main changes are:
* If there is `GJKSolver_[libccd/indep]::shapeIntersect<ShapeA, ShapeB>(~)` then add `GJKSolver_[libccd/indep]::shapeIntersect<ShapeB, ShapeA>`.
* For the added `GJKSolver_[libccd/indep]::shapeIntersect<ShapeB, ShapeA>` functions, the reverses are reversed.

scpeters:
:+1:

jslee02:
Waiting a day for further comments before merging

wjwwood:
This bug is still present in Indigo. It would be nice to backport these changes to 0.3.x and release a patch version (like 0.3.3) that could be put into ROS Indigo. If there is interest in doing this I would be happy to help.

jslee02:
I can backport these changes to (probably) 0.3.1, but might need help releasing a patch for ROS Indigo (actually don't know anything about it XD).

wjwwood:
It would be best to have an official new version (current latest is 0.3.2, so the new one would be 0.3.3). Otherwise, I could apply a patch in our release process, but that's not ideal.

I'm not sure how you want to do it, but what I would do is:

- make a `0.3` branch using the `0.3.2` tag (since `master` has moved on since then)
- cherry-pick these fixes to the new `0.3` branch
- make a new release with a new tag `0.3.3`

Once that's done I can make sure that tagged `0.3.3` version is released into Indigo for you.

wjwwood:
Sorry, there's already a `0.3` branch:

https://github.com/flexible-collision-library/fcl/tree/fcl-0.3

I missed it before somehow.

jslee02:
That plan sounds good to me. Will work on it. I just created the branch a min ago.

wjwwood:
Thanks!

jslee02:
PR created.

jslee02:
[FCL 0.3.3 is released](https://github.com/flexible-collision-library/fcl/releases/tag/0.3.3). You would be able to release the packages for Indigo now.

wjwwood:
Thanks for making the release so quickly! I've proposed version 0.3.3 for Indigo and Jade (both on Trusty, but also other systems):

- Indigo: https://github.com/ros/rosdistro/pull/12220
- Jade: https://github.com/ros/rosdistro/pull/12221

PR #56, opened 2015-04-05
Enable out-of-source build by generating configs to binary directory

jslee02:
One recommended feature of CMake is the ability to do out-of-source builds. However, FCL currently generates the config files (fcl.pc, include/fcl/config.h, test/fcl_resources/config.h) into the source directories, which makes it not possible to do out-of-source build. This pull request changes the output directory to the binary directory to resolve it.


scpeters:
:+1:

PR #58, opened 2015-04-06
Fix inconsistent normal directions

jslee02:
According to [the comment](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/collision_data.h#L80), contact normal should point from object 1 to object 2. However, shape-shape and shape-mesh collision algorithms doesn't follow the convention. This pull request fixes it.

A bunch of tests in test/test_fcl_geometric_shapes.cpp are updated to be able to check the normals.

### DIFF
--- a/include/fcl/math/transform.h
+++ b/include/fcl/math/transform.h
@@ -171,6 +171,12 @@ Quaternion3f conj(const Quaternion3f& q);
 /// @brief inverse of quaternion
 Quaternion3f inverse(const Quaternion3f& q);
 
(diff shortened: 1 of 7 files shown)


scpeters:
You changed thousands of lines in `test/test_fcl_geometric_shapes.cpp`. That diff is too large to review. Is there any way to make it smaller or split it into multiple commits?

jslee02:
It's possible. I'll split it into multiple commits.

jslee02:
This PR is replaced by #61.

PR #59, opened 2015-04-08
List headers in add_library for IDEs

jslee02:
Basically listing headers to `add_library` is not necessary to generate libraries and doesn't affect on the libraries. It essentially makes headers to show up in IDEs like Visual Studio, Xcode, or Qt Creator. It may be meaningless to the users don't use IDEs but I think it would be pretty useful to who uses IDEs.


scpeters:
:+1:

jslee02:
Waiting a day for further comments before merging

PR #60, opened 2015-04-09
Support Halfspace through main [collision/distance/continuous collision] interface

### LINKED ISSUE #57: Halfspace is not supported by main collision interface
`Halfspace` is one of primitive shapes of FCL but it's not registered to the collision matrix. So it can't be used through `collide(~)`, which is the main collision interface.

jslee02:
This pull request enables to use Halfspace through the main interface. FCL contains Halfspace shape and necessary algorithms for [collision/distance/continuous collision], but it couldn't be used through the main interface `collide(~)` because it's not registered to the corresponding function matrices. Here is an example code that prints warning messages and doesn't compute collision result for Halfspace.

```cpp
Cylinder s(5, 10);
Halfspace hs(Vec3f(1, 0, 0), 0);
Transform3f tf1;
Transform3f tf2;

CollisionRequest request;
CollisionResult result;

// Following code prints warning message:
// Warning: collision function between node type 13 and node type 16 is not supported
collide(&s, tf1, &hs, tf2, request, result);
```

This PR resolves #57.


jslee02:
It's fixed by 40a0caf.

scpeters:
:+1:

scpeters:
I didn't like the github diff colorization, so I used meld on my local machine, and it was much easier to see the changes. Nice work on generalizing the tests.

jslee02:
Agreed, it's hard to see the changes in here especially when the changes are more than 1k lines. The tests are still long, maybe I could shorten more once I get more used to Boost Test Library. 

I'll wait one another day for further comments.

PR #62, opened 2015-04-16
Update copyright information

jslee02:
This pull request adds LICENSE file to the top-level directory, and updates the copyright years. The range of copyright years of LICENSE covers from the most old year in the source code to 2015.

Things to do: I found that deprecated.h contains mixed copyright holders, CNRS-LAAS ([line 4](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/deprecated.h#L4)) and Willow Garage ([line 17](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/deprecated.h#L17)). I think we should use consistent copyright holder in one license text but don't know which one is the correct one is the original.


scpeters:
Looking at the [history of deprecated.h](https://github.com/flexible-collision-library/fcl/commits/master/include/fcl/deprecated.h), it looks like that file was only edited by @florent-lamiraux , so I would assume that `CNRS-LAAS` is the proper holder.

scpeters:
:+1:

jslee02:
It seems so. I'll merge this PR in one day waiting another comment.

florent-lamiraux:
I have copied the file from another project. Copyright is thus owned by CNRS and AIST.
The authors are Thomas Moulard (https://github.com/thomas-moulard) and Kayusawa (https://github.com/kayusawa) for the windows part.

mikeferguson:
Since willow garage no longer exists in it's former form -- most people have been assigning new work to the Open Source Robotics Foundation. It seems strange to me to have a 2015 copyright with Willow Garage in the name....

mikeferguson:
For instance, here is an example of adding new copyrighted work to OSRF: https://github.com/ros/roscpp_core/blob/ec54971462abdc94db390b1b8c285bc800b4ff1c/cpp_common/include/ros/header.h

jslee02:
Oh, I didn't noticed that. Since Willow Garage closed Jan 2014, I think it should be like (but not sure):
```
Copyright (c) 2008-2014, Willow Garage, Inc.
Copyright (c) 2014-2015, Open Source Robotics Foundation
```

Also, the third phrase seems to be updated like:
```
 *   * Neither the name of Open Source Robotics Foundation, Inc. nor the names
 *     of its contributors may be used to endorse or promote products derived
 *     from this software without specific prior written permission.
```

scpeters:
Good points, splitting the dates seems reasonable to me.

jslee02:
Regarding the holders of deprecated.h and some other files, 
```
Copyright (c) 20xx-2015, CNRS-LAAS and AIST
```
would be the correct one? or They should be the full names of them?

florent-lamiraux:
CNRS-LAAS and AIST is perfect.

jslee02:
Updated copyright information based on above suggestions.

jslee02:
I believe it's ready to merge. I will wait one another day for further comment.

PR #63, opened 2015-04-16
Fix box-box intersection algorithm with test

jslee02:
I recreated pull request #42 adding some tests. It wasn't able to add the tests into the original pull request since I don't have access to commit the author's forked repository.

It is confirmed that the tests [fail](https://travis-ci.org/dartsim/fcl/builds/58801002) without the fix.


src/narrowphase/narrowphase.cpp
```diff
@@ -1492,14 +1503,19 @@ bool boxBoxIntersect(const Box& s1, const Transform3f& tf1,
...
+    // for(size_t i = 0; i < contacts.size(); ++i)
+    // {
+    //   contact_point += contacts[i].point;
+    // }
 
-    contact_point = contact_point / (FCL_REAL)contacts.size();
+    // contact_point = contact_point / (FCL_REAL)contacts.size();
 
-    *contact_points = contact_point;
+    // *contact_points = contact_point;
```
scpeters: Wouldn't it be better to delete this old code than leave it mysteriously commented?

scpeters:
Can you resolve the conflicts?

src/narrowphase/narrowphase.cpp
jslee02: Sure, removed by 366a915. I guess the original author (@EdsterG) would think so as well.

jslee02:
Resolved. 

As a side note, the box-box collision algorithm needs to be changed to return multiple contact points instead of averaging them once #52 is merged.

scpeters:
:+1:

Issue #41, opened 2014-11-25
Segfault when calling DynamicAABBTreeCollisionManager::registerObjects with empty vector


Zera-:
I am trying to integrate FCL into an existing scenegraph-system, that internally manages multiple levels of object-grouping. In order to use the broadphase algorithms of FCL I fill instances of fcl::DynamicAABBTreeCollisionManager for each grouping.

During the creation of a higher-level grouping I then attempt to add the list of collision objects in the subgroup that is currently empty.

The manager starts a tree traversal for insertion, and the final insertion method HierarchyTree<BV>::topdown_0(in file include/fcl/broadphase/hierarchy_tree.hxx, line 353) then tries to touch the first element of the empty list(line 386), causing a segfault.

The method topdown_0 even checks the number of elements before doing the main processing, so I assume it could be a reasonable place to catch the special case of an empty list. However, I can't really assess the affects of setting the root pointer of the tree to NULL (as would be my natural answer to no insertion).

I suspect that similar cases are also currently unhandled by the library, making the usage quite unrelieable.

If you could set a guideline on when special cases should be caught (As early as possible? Or process them until they have to be handled?), I'd be up to fixing issues I find myself and submitting pull requests.

jslee02:
A pull request #64 was created to solve this issue.

jslee02:
I believe this issue is fixed by #64.

PR #64, opened 2015-04-16
Fix segfault when registering empty object vector to [~]CollisionManager

jslee02:
Three broad-phase collision managers (`DynamicAABBTreeCollisionManager`, `DynamicAABBTreeCollisionManager_Array`, and `SaPCollisionManager`) don't take into account a case of attempting to register collision objects with empty vector, which is also reported by #41. This pull request fixes it.

### DIFF
--- a/test/test_fcl_broadphase.cpp
+++ b/test/test_fcl_broadphase.cpp
@@ -149,6 +149,26 @@ BOOST_AUTO_TEST_CASE(test_core_bf_broad_phase_self_distance)
   broad_phase_self_distance_test(200, 5000);
 }
 
+/// check broad phase collision for empty collision object set and quaries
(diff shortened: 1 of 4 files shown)


test/test_fcl_broadphase.cpp
```diff
@@ -149,6 +149,26 @@ BOOST_AUTO_TEST_CASE(test_core_bf_broad_phase_self_distance)
   broad_phase_self_distance_test(200, 5000);
 }
 
+/// check broad phase collision for empty collision object set and quaries
```
scpeters: quaries?

scpeters:
:+1: aside from spelling of queries

jslee02:
Fixed the typo, thanks. I'll merge this PR in one day waiting another comment.

Issue #23, opened 2014-01-10
error in MSVC


JunghwanLee:
'__declspec' should be placed before the function definition in MSVC.

jslee02:
I believe this issue is fixed by #65.

PR #65, opened 2015-04-25
Place of FCL_DEPRECATED for portability

jslee02:
This PR resolves issue #23.


scpeters:
@slater1 how does this look?

scpeters:
+1

PR #68, opened 2015-05-03
Fix incorrect state check in BVHModel::addVertex(~)

### LINKED ISSUE #20: Failure in building mesh using fcl::BVHModel< fcl::OBB >::addVectex
I try to build a mesh using the following naive procedure
```
boost::shared_ptr <fcl::BVHModel< fcl::OBB > mesh
(new (fcl::BVHModel< fcl::OBB >));
mesh->beginModel ();
// Loop over a previously filled vector of vectices
for (std::size_type  i = 0; i<vertices.size (); ++i)
  mesh->addVertex (vertices [i]);
// Loop over a previously filled vector of triangles
for (std::size_type  i = 0; i<triangles.size (); ++i)
  mesh->addVertex (triangles [i]);
mesh->endModel ();
```
This does not works since method beginModel sets ```build_state``` member of mesh
to ```BVH_BUILD_STATE_BEGUN``` and method ```addVertex``` expects BVH_BUILD_STATE_UPDATE_BEGUN as value for ```build_state```.

I noticed that addVertex is not used in tests. Did I do something wrong ?

jslee02:
This PR fixes #20.

### DIFF
--- a/include/fcl/math/transform.h
+++ b/include/fcl/math/transform.h
@@ -172,6 +172,12 @@ Quaternion3f conj(const Quaternion3f& q);
 /// @brief inverse of quaternion
 Quaternion3f inverse(const Quaternion3f& q);
 
+static inline std::ostream& operator << (std::ostream& o, const Quaternion3f& q)
+{
+  o << "(" << q[0] << " " << q[1] << " " << q[2] << " " << q[3] << ")";
+  return o;
+}
+
 
 /// @brief Simple transform class used locally by InterpMotion
 class Transform3f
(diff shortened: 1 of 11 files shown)


scpeters:
I manually applied the patch in e6ac89477ad4d5cecb82432ac02fe79d4aad018c to show that the test failed without the fix, then recompiled with the fix and it passes.

:+1: 

For extra credit from me, commit the test first and post travis results showing that the test fails without the fix, then passes after. Great work.

jslee02:
Thanks. The advise sounds reasonable. Will do so.

PR #69, opened 2015-05-26
Fix warning about abs() with float's

scpeters:
abs() only works for integers,
replace with fabs() for floaing point

Otherwise OS X has lots of complaints


jslee02:
I heard that `abs()` also works for floating point [[1]](http://en.cppreference.com/w/cpp/numeric/math/fabs)[[2]](http://stackoverflow.com/questions/3118165/when-do-i-use-fabs-and-when-is-it-sufficient-to-use-stdabs) in C++, but I haven't tested it on OS X. Testing on OS X...

jslee02:
It seems `abs()` directly refers to `int abs(int)` while `std::abs()` is overloaded for all the numerical types so it refers to `fabs()` for floating point. Explicitly using C function `fabs()` for floating point would also suppress the warnings, but I'm a little bit inclined to specify `std` namespace since it's C++.

scpeters:
Ok, I replaced `fabs` with `std::abs`. I used fabs because I wasn't sure if `std::abs` was a c++11 thing, but it doesn't appear to be so.

jslee02:
:+1:

Issue #67, opened 2015-05-03
Failure in building point cloud models for OBB/RSS/kIOS/OBBRSS


jslee02:
FCL causes segfault in building point cloud model as:

```cpp
boost::shared_ptr<BVHModel<OBB> > model(new BVHModel<OBB>);  // or RSS/kIOS<16/18/24>/OBBRSS
model->beginModel();
// Loop over a previously filled vector of vectices
for (std::size_t i = 0; i < points.size(); ++i)
  result = model->addVertex(points[i]);
model->endModel();  // segfault!
```

The main reason of the segfault is that the bvh models don't handle triangle indices correctly.

Also, some tests for point cloud models are needed since there is none.

ricardoglc:
It seems this is caused by an error in BVHModel<BV>::endModel(). The member tri_indices is supposed to be NULL when point clouds are used, however it's being initialized even when num_tris is zero.

ricardoglc:
I've submitted a patch, but even though it solves that specific problem, those BVHs still cannot be used for point cloud collision checking. There is no support for mesh x shape collision with point clouds as you can see in traversal_node_setup.h:

  if(model2.getModelType() != BVH_MODEL_TRIANGLES)
    return false;

PR #72, opened 2015-06-09
Fix failure in building point cloud models for OBB/RSS/kIOS/OBBRSS (issue #67)

ricardoglc:
The functions beginModel, addTriangle and addSubModel now respect documentation, keeping tri_indices as NULL for point clouds. This fixes issue #67 .

### DIFF
--- a/include/fcl/octree.h
+++ b/include/fcl/octree.h
@@ -54,7 +54,7 @@ namespace fcl
 class OcTree : public CollisionGeometry
 {
 private:
-  boost::shared_ptr<const octomap::OcTree> tree;
+  boost::shared_ptr<octomap::OcTree> tree;
 
   FCL_REAL default_occupancy;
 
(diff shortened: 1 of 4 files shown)


scpeters:
+1 for the test passing, but I'm not very familiar with the details of what has changed, so I'll wait for someone else to review as well.

include/fcl/octree.h
```diff
@@ -97,6 +97,12 @@ class OcTree : public CollisionGeometry
     aabb_radius = (aabb_local.min_ - aabb_center).length();
   }
 
+  /// @brief get the underlying octree structure
+  boost::shared_ptr<octomap::OcTree> getTree()
+  {
+    return tree;
+  }
```
jslee02: It seems this accessor is added to address #73, which looks good to me. It would also good to provide const version of `getTree()` for const correctness as:
```cpp
boost::shared_ptr<const octomap::OcTree> getTree() const
{
  return tree;
}
```

jslee02: We might be able to add this later.

jslee02:
:+1: I believe this pull request is ready to merge.

jslee02:
I'd like to merge this unless someone has an objection. @ricardoglc could resolve the confliction so that we could merge this?

jslee02:
Close in favor of #237

PR #74, opened 2015-07-17
Fixing missing assignment in Convex class

uschwes:
From what I understand the Convex class is not really supported, yet the missing assignment is an error as it is used uninitialized.

### DIFF
--- a/include/fcl/shape/geometric_shapes.h
+++ b/include/fcl/shape/geometric_shapes.h
@@ -278,6 +278,7 @@ public:
     plane_dis = plane_dis_;
     num_planes = num_planes_;
     points = points_;
+    num_points = num_points_;
     polygons = polygons_;
     edges = NULL;
 


include/fcl/shape/geometric_shapes.h
```diff
@@ -278,6 +278,7 @@ class Convex : public ShapeBase
     plane_dis = plane_dis_;
     num_planes = num_planes_;
     points = points_;
+    num_points = num_points_;
```
isucan: How did this ever work without this fix?

uschwes: Did it? I do not see any unit tests related to the ```Convex``` shape...

k-okada:
@jslee02, hi, can we backport htis to 0.3 so that ros users can use convex calculation?
https://github.com/flexible-collision-library/fcl/pull/141

jslee02:
@k-okada I think it's possible. Could you send a pull request for it? I might not able to to this right now.

PR #84, opened 2015-12-23
Make GJK struct variables static

gdbaldw:
Hi, The broadphase manager appears to use hardcoded GJK defaults for narowphase. I'm using broadphase, and have a need for a smaller collision tolerance in narowphase. In my main program, I can set the narowphase default with,
```
fcl::GJKSolver_libccd::collision_tolerance = 1e-15;
```
All the other defaults can be reset as well with this pull requested code.
-Doug Baldwin

### DIFF
--- a/include/fcl/narrowphase/narrowphase.h
+++ b/include/fcl/narrowphase/narrowphase.h
@@ -203,10 +203,6 @@ struct GJKSolver_libccd
   /// @brief default setting for GJK algorithm
   GJKSolver_libccd()
   {
-    max_collision_iterations = 500;
-    max_distance_iterations = 1000;
-    collision_tolerance = 1e-6;
-    distance_tolerance = 1e-6;
   }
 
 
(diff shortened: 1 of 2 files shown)


isucan:
I am not sure this is a good idea. This makes it such that all instances have to use the same config options, which is not idea. I do not recommend merging this.

jslee02:
I agree with @isucan in principal, but currently, there is no chance to change the values anyways when we use broadphase manager or [the main collision interface](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/collision.h#L49) because `GJKSolver_libccd` (and also `GJKSolver_indep`) is [used as local variable](https://github.com/flexible-collision-library/fcl/blob/master/src/collision.cpp#L126).

I guess @gdbaldw wanted to change the values by making them as static members.

One possible compromise I can come up with now is that we pass the values in `CollisionRequest`, that was designed for delivering collision options. To do so, we might need to make `CollisionRequest` and `collision()` as templated struct and function, respectively.

gdbaldw:
Yes @jslee02. I have no urgency, so a more thoughtful architectural approach using CollisionRequest is nice too. I share @isucan's concern, and offered this as a quick solution to highlight the issue.

PR #85, opened 2016-01-01
Use find_library for library search instead of find_path

jslee02:
`find_library` is more robust than `find_path` for library search because `find_library` considers the library name as the library file name and also considers with platform-specific prefixes (e.g. `lib`) and suffixes (e.g. `.so`) while `find_path` doesn't.


scpeters:
+1

PR #87, opened 2016-01-05
SphereSphereIntersect point bugfix

gdbaldw:
The bug is that for sphere objects of differing radii, the collision point is dependent on the order of the objects, and is a non-linear function. With this proposed fix, the collision point is independent of the order of the sphere objects, and is a linear function. A comparison test of (o1, o2) with (o2, o1) would have identical pos.

This bug and fix is indicative of a pervasive set of fcl bugs when using a precise dynamic solver. Another example is sphere-plane, which again has a different pos depending on the order (o1, o2) or (o2, o1). In that case, pos would need to be the midpoint of the intersected radius, not the extreme.

In general, for consistent input to a precise dynamic solver, I will be making changes to fcl such that pos is always independent of object order. This means that for box-plane, pos will be at the centroid (I believe) of the intersection. In this way, pos will move incrementally over time instead of jumping from corner to corner of the box as can happen in the current implementation.

The solver I'm using is MBDyn, which fits perfectly with fcl, once a consistent interface is achieve as indicated by the above comments. I believe this kind of consistent interface has merit beyond my needs. Please let me know if you would like such changes pushed for consideration to merge.

### DIFF
--- a/src/narrowphase/narrowphase.cpp
+++ b/src/narrowphase/narrowphase.cpp
@@ -278,9 +278,9 @@ bool sphereSphereIntersect(const Sphere& s1, const Transform3f& tf1,
   {
     // If the centers of two sphere are at the same position, the normal is (0, 0, 0).
     // Otherwise, normal is pointing from center of object 1 to center of object 2
-    const Vec3f normal = len > 0 ? diff / len : diff;
-    const Vec3f point = tf1.transform(Vec3f()) + diff * s1.radius / (s1.radius + s2.radius);
+    const Vec3f normal = len > 0 ? diff / len : Vec3f();
     const FCL_REAL penetration_depth = s1.radius + s2.radius - len;
+    const Vec3f point = tf1.transform(Vec3f()) + normal * (s1.radius - 0.5 * penetration_depth);
     contacts->push_back(ContactPoint(normal, point, penetration_depth));
   }
 
--- a/test/test_fcl_geometric_shapes.cpp
+++ b/test/test_fcl_geometric_shapes.cpp
@@ -523,7 +523,7 @@ BOOST_AUTO_TEST_CASE(shapeIntersection_spheresphere)
   tf2 = Transform3f(Vec3f(29.9, 0, 0));
   contacts.resize(1);
   contacts[0].normal.setValue(1, 0, 0);
-  contacts[0].pos.setValue(20.0 - 0.1 * 20.0/(20.0 + 10.0), 0, 0);
+  contacts[0].pos.setValue(20.0 - 0.5 * 0.1, 0, 0);
   contacts[0].penetration_depth = 0.1;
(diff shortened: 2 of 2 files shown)


scpeters:
+1

jslee02:
I agree that collision algorithms should be independent on the order of the objects, but I'm not sure the current `SphereSphereIntersect` is not. 

It returns the point of internal divisions corresponding to the radii rather than mid point [since this](https://github.com/flexible-collision-library/fcl/commit/0b8fcac739a87bd0242e3c94fefdbdb0c596bb17). However, it is still independent on the order of objects. Please check [these tests](af165d9e2b80164e4d5609ccf99de4deaea98ee0) -- [passing the tests](http://clicks.travis-ci.org/track/click/30007208/travis-ci.org?p=eyJzIjoiRkNBd0xwYU0xc0JXVFoxbVNqenVYcWJjZVZzIiwidiI6MSwicCI6IntcInVcIjozMDAwNzIwOCxcInZcIjoxLFwidXJsXCI6XCJodHRwczpcXFwvXFxcL3RyYXZpcy1jaS5vcmdcXFwvZGFydHNpbVxcXC9mY2xcXFwvYnVpbGRzXFxcLzEwMTI0NTgzMlwiLFwiaWRcIjpcImNmNmYwNTc2NDEwNzRiZTliNTJiNGIxZTFiOGZlMTQyXCIsXCJ1cmxfaWRzXCI6W1wiNWNhNzZhYzk3ZDliY2FhNjdjYjExZjRlOGRiNDdjZjc5NDg0OWE0YlwiXX0ifQ).

gdbaldw:
@jslee02, thank you for correcting me. (I should have written the test first myself. The bug was in my object swapping code.) Regarding the larger topic, I'd like to have a conversation, if you would allow.

1) Is there a consistent general fcl-philosophy for calculating contact point, depth, and normal? (e.g. box-plane compared to sphere-plane).

2) Could fcl-master optionally include intersect points? For example, rolling friction of sphere-plane can be simulated with four additional contact points at the circular surface intersection,  x=Circle_Center_x and y=Circle_Center_y.

3) I think my general philosophy would be to have one contact at the intersecting solid centroid (using this algorithm for hulls, http://wwwf.imperial.ac.uk/~rn/centroid.pdf), with the normal plane defined by the intersecting solid's major and minor axes. And, have four additional contacts at the intersecting surface based on these major and minor axes. Does this seem reasonable?

I understand many fcl applications require computational speed, and added contacts could be a burden. I do not yet fully understand fcl, as I am just diving into learning contact methods. I'm striving for higher levels of dynamic simulation precision, so believe I will need a fork. But, if successful and coordinated, this fork could eventually merge, and offer configurable options within fcl-master.

I would appreciate your insight into these questions, as time may permit you to offer guidance.

-Doug

PR #89, opened 2016-01-08
Use GNUInstallDirs to set install directories

jpgr87:
Use the GNUInstallDirs CMake script to set the installation directories
for shared libraries, includes, and pkgconfig files.  Allows for library
installation directories other than "lib" for distribution packaging. (e.g. lib/x86_64-linux-gnu/
to support debian/ubuntu multiarch, and lib64/ to support fedora/redhat multiarch)


jslee02:
I don't know about `GNUInstallDirs` yet, but we should increase the minimum required cmake version to at least 2.8.5 to use the script since it's supported by cmake 2.8.5 or greater.

jpgr87:
That's a good point, I pushed a commit with that update.

src/CMakeLists.txt
```diff
@@ -7,4 +7,7 @@ endif()
 
 target_link_libraries(${PROJECT_NAME} ${CCD_LIBRARIES} ${OCTOMAP_LIBRARIES} ${Boost_LIBRARIES})
 
-install(TARGETS ${PROJECT_NAME} ARCHIVE DESTINATION lib/ LIBRARY DESTINATION lib/)
+install(TARGETS ${PROJECT_NAME}
+  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
```
jslee02: It seems `RUNTIME` is not necessary as fcl doesn't install executables.

jpgr87: It's not necessary on linux systems, but on windows systems the runtime destination is used to determine where to install the .dll.  I usually add all three out of habit, if you're not worried about supporting windows systems I can remove the runtime component.

jslee02: I didn't consider Windows habitually. :disappointed: Since fcl supports Windows, we need the runtime component.

jslee02:
:+1:  Looks good to me.

src/CMakeLists.txt
jpgr87: No worries.  It's more important for the case where you are shipping a mix of dlls and exes linked to them.  Specifying that the dlls belong in the bindir makes it so that they are in the same directory as executables, which lets you run the executables without having to edit your path environment variable (the current directory is the windows' first library search path.)

PR #94, opened 2016-02-23
Fix build when libccd package config not found.

law12019:
Link to ccd library by full path per preferred cmake convention.  This fixed the build when libccd is not found by pkgconfig.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,10 +129,10 @@ if(NOT CCD_FOUND)
     # of ccd
     find_path(CCD_INCLUDE_DIRS ccd.h
         PATH_SUFFIXES ccd)
-    find_library(CCD_LIBRARY_DIRS
+    find_library(CCD_LIBRARY
         ${CMAKE_SHARED_LIBRARY_PREFIX}ccd${CMAKE_SHARED_LIBRARY_SUFFIX})
-    if(CCD_INCLUDE_DIRS AND CCD_LIBRARY_DIRS)
-        set(CCD_LIBRARIES "ccd")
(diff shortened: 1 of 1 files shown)


scpeters:
What do you think @j-rivero ?

j-rivero:
> What do you think @j-rivero ?

+1

mamoll:
In a similar vain, but using pkg-config, I use a [FindCCD.cmake](https://github.com/ompl/omplapp/blob/master/CMakeModules/FindCCD.cmake) module in a project that depends on both ccd and fcl:
```
include(FindPackageHandleStandardArgs)
find_package(PkgConfig)
if(PKGCONFIG_FOUND)
    pkg_check_modules(CCD ccd>=2.0)
    if(CCD_LIBRARIES AND NOT CCD_INCLUDE_DIRS)
        set(CCD_INCLUDE_DIRS "/usr/include")
    endif()
endif()
find_package_handle_standard_args(CCD DEFAULT_MSG CCD_LIBRARIES CCD_INCLUDE_DIRS)
```
This has the added advantage that you can check for the appropriate libccd version.

sherm1:
Is this ready to merge?

mamoll:
:+1:

j-rivero:
+1

jslee02:
I would prefer to use `FindCCD.cmake` as @mamoll suggested, but this looks good as a stopgap solution. Also, adding `CCD_VERSION` variable would be considerable to add to `FindCCD.cmake` as [this](https://github.com/dartsim/dart/blob/master/cmake/FindCCD.cmake#L33).

PR #99, opened 2016-03-21
Small fixes to allow compilation in Visual Studio 2015, and get rid of some warnings

sherm1:
I tried building FCL on Windows 10 with Visual Studio 2015 Win64 and got a couple of compilation failures (there was a missing include and use of non-standard constant `M_PI_2`). This PR fixes those and allows FCL to build. However 4 or 13 tests failed -- those need to be addressed but I'm not attempting that in this PR.

There were also a few warnings when I built in Ubuntu 14.04; those are fixed here too. All tests passed there.

### DIFF
--- a/test/test_fcl_capsule_capsule.cpp
+++ b/test/test_fcl_capsule_capsule.cpp
@@ -37,6 +37,7 @@
 
 #define BOOST_TEST_MODULE "FCL_CAPSULE_CAPSULE"
 #include <boost/test/unit_test.hpp>
+#include <boost/math/constants/constants.hpp>
 
 #include "fcl/collision.h"
 #include "fcl/shape/geometric_shapes.h"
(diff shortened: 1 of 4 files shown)


test/test_fcl_capsule_capsule.cpp
```diff
@@ -126,6 +125,7 @@ BOOST_AUTO_TEST_CASE(distance_capsulecapsule_transformZ)
 
 BOOST_AUTO_TEST_CASE(distance_capsulecapsule_transformZ2)
 {
+  const FCL_REAL Pi = std::acos(FCL_REAL(-1.0));
```
isucan: I would prefer either using boost::math::constants::pi<double> or defining M_PI if not defined.
{{{
#ifndef M_PI
#define M_PI 3.1415926535....
#endif
}}}

sherm1: I was trying to fix what appeared to be incorrect use of floating point type. Using macros for constants means they will be untyped and might not convert properly to FCL_REAL. I took a look at boost::pi and can't figure out what type it is! I don't think it is a good idea to have the digits of pi included here -- there must be a better way! What would you suggest as the proper way to get a "pi" of type FCL_REAL?

sherm1: Oh, I see -- it's a template function. I'll use it.

sherm1:
@isucan -- I replaced the calculated pi with the boost version.

isucan:
Thanks!

PR #104, opened 2016-03-23
replace much boost usage with C++11 equivalents

mamoll:
A first step in eliminating Boost usage. Mostly low-hanging fruit.


src/shape/geometric_shapes_utility.cpp
```diff
@@ -94,10 +94,7 @@ std::vector<Vec3f> getBoundVertices(const Ellipsoid& ellipsoid, const Transform3
 
   std::vector<Vec3f> result(12);
 
-  const FCL_REAL phi = (1.0 + sqrt(5.0)) / 2.0;  // golden ratio
-  // TODO: Replace with constexpr when we migrate to C++11 or
-  //       boost::math::constants::phi<FCL_REAL>() if boost version is greater
-  //       than 1.50.
+  const FCL_REAL phi = (1.0 + std::sqrt(5.0)) / 2.0;  // golden ratio
```
jslee02: Can we consider using `constexpr` for `phi`, `a`, and `b` as we're migrating to using C++11?

mamoll: Yes, but sqrt is not allowed in a `constexpr` (at least not with latest Xcode).

include/fcl/BV/RSS.h
```diff
@@ -38,10 +38,10 @@
 #ifndef FCL_RSS_H
 #define FCL_RSS_H
 
-
+#include <math.h>
```
jslee02: nitpick: I would prefer to use `<cmath>` since this is a C++ standard library header while `<math.h>` is a header defined by the C standard library.

isucan:
+1 for merging.

include/fcl/BV/RSS.h
sherm1: :+1: for `#include <cmath>`. Shouldn't use any C headers.

sherm1:
I'll test this on Windows since CI doesn't. Also, Travis failed.

mamoll:
Travis is failing probably because of ancient compilers: gcc 4.6 and clang 3.4. I guess the Travis build matrix needs to be updated. We should also set up Windows builds on Appveyor. I'll make the <cmath> change. I'll wait a bit to give others a chance to respond.

include/fcl/BV/RSS.h
```diff
@@ -111,7 +111,7 @@ class RSS
   /// @brief Volume of the RSS
   inline FCL_REAL volume() const
   {
-    return (l[0] * l[1] * 2 * r + 4 * boost::math::constants::pi<FCL_REAL>() * r * r * r);
+    return (l[0] * l[1] * 2 * r + 4 * M_PI * r * r * r);
```
sherm1: Two problems with `M_PI`:
- it is a macro and hence untyped; needs to have type FCL_REAL
- it is non-standard (it is a Posix feature, not a C++11 one)

Since "pi" is the only boost math constant FCL used, I think the simplest thing would be to define an fcl::kPi or whatever that could be defined in a common header with something like
```
constexpr FCL_REAL kPi=FCL_REAL(/*lots of digits of pi copied from a reliable source*/);
```

src/shape/geometric_shapes_utility.cpp
jslee02: I found that none of the math functions in C++11 are not allowed to use with `constexpr` (gcc allows it, but [it seems not to be allowed in C++14](http://stackoverflow.com/a/27744080)). Okay, we can't use `constexpr` here.

jslee02:
> Travis is failing probably because of ancient compilers: gcc 4.6 and clang 3.4.

This can be addressed by using [trusty environment of Travis](https://docs.travis-ci.com/user/trusty-ci-environment/).

src/broadphase/broadphase_interval_tree.cpp
```diff
@@ -178,9 +179,9 @@ void IntervalTreeCollisionManager::setup()
 {
   if(!setup_)
   {
-    std::sort(endpoints[0].begin(), endpoints[0].end(), boost::bind(&EndPoint::value, _1) < boost::bind(&EndPoint::value, _2));
-    std::sort(endpoints[1].begin(), endpoints[1].end(), boost::bind(&EndPoint::value, _1) < boost::bind(&EndPoint::value, _2));
-    std::sort(endpoints[2].begin(), endpoints[2].end(), boost::bind(&EndPoint::value, _1) < boost::bind(&EndPoint::value, _2));
+    std::sort(endpoints[0].begin(), endpoints[0].end());
+    std::sort(endpoints[1].begin(), endpoints[1].end());
+    std::sort(endpoints[2].begin(), endpoints[2].end());
```
sherm1: That looks a lot nicer in C++11 than in boost!

mamoll: This is actually not a C++11-related change. (Well, sort of, s/boost/std/g didn't work.) The simplification is because of the operator< declaration in the Endpoint class.

src/shape/geometric_shapes_utility.cpp
sherm1: Could do it constexpr by precalculation (as I suggested for pi) if you can find a very high-precision calculator.

src/shape/geometric_shapes_utility.cpp
```diff
@@ -94,10 +94,7 @@ std::vector<Vec3f> getBoundVertices(const Ellipsoid& ellipsoid, const Transform3
...
 
   const FCL_REAL a = std::sqrt(3.0) / (phi * phi);
```
sherm1: There is an implication here that FCL_REAL has no higher precision than double (that is, it can't be long double), because we're initializing it with a double sqrt(3).  I'm fine with that limitation but we ought to say so.

sherm1:
This is great, Mark -- it's so nice to see all those boosts disappear! I hope it still works -- the test suite seems kind of short.

mamoll:
Does anyone object to defining all mathematical constants (pi and phi) in, say, include/fcl/math/constants.h in the fcl::constants namespace as hard-coded double values:
```
namespace fcl
{
   namespace constants
   {
       constexpr FCL_REAL pi = 3.14159....;
       constexpr FCL_REAL phi = 1.618....;
   }
}
```

isucan:
I think that's a good idea!

jslee02:
I was considering letting the compiler do the calculation of `phi` in compile like [this](https://github.com/pkeir/ctfft/blob/master/const_math.hpp#L22), but @mamoll 's idea looks much simpler!

CMakeModules/CompilerSettings.cmake
```diff
@@ -1,3 +1,6 @@
+# force C++11 mode
+add_definitions(-std=c++11)
```
sherm1: This causes zillions of warnings on Windows -- that should only be added for gcc & clang.

CMakeModules/CompilerSettings.cmake
```diff
@@ -1,3 +1,6 @@
...
+
 if(CMAKE_COMPILER_IS_GNUCXX)
     add_definitions(-W -Wall -g -Wextra -Wno-missing-field-initializers -Wno-unused-parameter)
```
sherm1: I think you could just move the -std=c++11 here also.

include/fcl/BV/RSS.h
sherm1: M_PI caused compilation failures in Visual Studio 2015.

CMakeModules/CompilerSettings.cmake
scpeters: :+1:

sherm1:
The top-level CMakeLists.txt specifies boost components [here](https://github.com/flexible-collision-library/fcl/blob/master/CMakeLists.txt#L107). Can you remove any of those with this PR?

include/fcl/BV/RSS.h
scpeters: `M_PI` works with visual studio if you define `_USE_MATH_DEFINES` before including `cmath`

~~~
#ifndef _USE_MATH_DEFINES
# define _USE_MATH_DEFINES
#endif
~~~

include/fcl/BVH/BVH_model.h
```diff
@@ -268,10 +268,10 @@ class BVHModel : public CollisionGeometry,
   BVHBuildState build_state;
 
   /// @brief Split rule to split one BV node into two children
-  boost::shared_ptr<BVSplitterBase<BV> > bv_splitter;
+  std::shared_ptr<BVSplitterBase<BV> > bv_splitter;
 
   /// @brief Fitting rule to fit a BV node to a set of geometry primitives
-  boost::shared_ptr<BVFitterBase<BV> > bv_fitter;
+  std::shared_ptr<BVFitterBase<BV> > bv_fitter;
```
scpeters: this is a change to the public API, should we bump the major version number?

scpeters:
I personally prefer separating big changes into smaller pull requests, but I'll defer to the crowd here.

include/fcl/profile.h
```diff
@@ -58,7 +58,7 @@
 #include <map>
 #include <string>
 #include <iostream>
-#include <boost/thread.hpp>
+#include <thread>
```
sherm1: Add `#include <mutex>` here -- compile failed on Windows. C++11 doesn't promise that mutex will be defined in thread.

include/fcl/BVH/BVH_model.h
jslee02: As far as I know, we might don't need to bump the major version number because FCL is still in 0.x.x.

sherm1:
With these changes, this PR builds on Windows VS 2015 Win64, with only the same 4 test failures I saw before:
- defined pi
- commented out "-std=c++11"
- included `<mutex>`

include/fcl/BVH/BVH_model.h
scpeters: good point, we should at least bump the minor version then

sherm1:
I like the constants proposal! I suggest using higher-precision constants that will get rounded down to FCL_REAL, to make sure the last bit is correct. For example:
```
constexpr FCL_REAL pi = FCL_REAL(3.141592653589793238462643383279502884197169399375105820974944592L);
```
That is a `long double` constant because of the trailing L.

include/fcl/BV/RSS.h
sherm1: > M_PI works with visual studio if you define _USE_MATH_DEFINES

Thanks -- I know about that but it's a non-standard hack so I think Mark's `fcl::constants::pi` idea is better.

scpeters: sounds good

mamoll:
I have hopefully addressed the main concerns. After this is merged, I'll fix the Travis CI config.

scpeters:
There's still packaging issues:

- [ ] bump minor version number
- [ ] export c++11 flags in pkgconfig / cmake config files

scpeters:
I guess those can be addressed later too?

mamoll:
I can fix the pkgconfig / cmake config files, but wouldn't it make more sense to bump the version when you actually release? There's still quite a bit of boost stuff to eliminate.

mamoll:
Actually, what cmake config files are you talking about? I thought you meant files like fcl-config.cmake, but this is currently not generated by FCL.

scpeters:
You're right; there is no cmake config file. I just assumed this had one.

I like to bump the version number in the master branch (CMakeModules/FCLVersion.cmake) when breaking changes happen. This way people building from source can see a difference by the version number. I consider the package released when a new tag has been created.

scpeters:
Thanks!

jslee02:
> export c++11 flags in pkgconfig / cmake config files

Actually, FCL has a [pkgconfig file](https://github.com/flexible-collision-library/fcl/blob/master/fcl.pc.in). Can we simply add `-std=c++11` like [gazebo does](https://bitbucket.org/osrf/gazebo/src/d17d941637690ecf5586afd05ac581da7f515936/cmake/pkgconfig/gazebo.in?at=default&fileviewer=file-view-default#gazebo.in-10)?

mamoll:
Apparently, MSVC doesn't like `-std=c++11`. This is why I add the flag only when not building with MSVC.

sherm1:
> I have hopefully addressed the main concerns. After this is merged, I'll fix the Travis CI config.

Mark, can you fix Travis as part of this PR so we can confirm that it passes? I think that just means changing `.travis.yml` as a commit with the PR.

sherm1:
Reviewed changes -- looks great! Ready to merge pending build success.

scpeters:
I think adding the following to `.travis.yml` would fix it:

~~~
sudo: required
dist: trusty
~~~

since trusty has a c++11 compiler

PR #105, opened 2016-03-24
Remove more boost usage and replace with C++11 equivalents.

mamoll:
Note: BVH_Model has a copy constructor, so having it derive from boost::noncopyable makes no sense. We can delete the copy assignment operator, if desired.
Only remaining boost usage in include/ and src/ is boost::dynamic_bitset. Use std::vector\<bool\> instead?

### DIFF
--- a/src/profile.cpp
+++ b/src/profile.cpp
@@ -222,7 +222,7 @@ void fcl::tools::Profiler::printThreadInfo(std::ostream &out, const PerThread &d
   {
     const AvgInfo &a = data.avg.find(avg[i].name)->second;
     out << avg[i].name << ": " << avg[i].value << " (stddev = " <<
-      sqrt(std::fabs(a.totalSqr - (double)a.parts * avg[i].value * avg[i].value) / ((double)a.parts - 1.)) << ")" << std::endl;
+      std::sqrt(std::abs(a.totalSqr - (double)a.parts * avg[i].value * avg[i].value) / ((double)a.parts - 1.)) << ")" << std::endl;
   }
 
   std::vector<dataDoubleVal> time;
--- a/src/profile.cpp
+++ b/src/profile.cpp
@@ -37,7 +37,7 @@
 /** \author Ioan Sucan */
 
 #include "fcl/profile.h"
-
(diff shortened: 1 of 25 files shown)


sherm1:
This is great, Mark! I'll try it on Windows.

sherm1:
> Only remaining boost usage in include/ and src/ is boost::dynamic_bitset. Use std::vector<bool> instead?

It would be easier if it could be replaced by `std::bitset` although that requires a compile-time size. However, it looks to me as though it is only used as a specialization of morton_functor, which requires the size to be provided at construction and then doesn't allow that size to be changed dynamically. And then that specialization isn't used in FCL except in a test case, where a constant size is used. The bitset specialization of morton_functor could make `bit_num` a template argument.

Does anyone know if there is an actual requirement for a dynamically-sized morton code? Jia, do you remember why this is using dynamic_bitset and whether that is required? @panjia1983

isucan:
+1 for merging!

sherm1:
Works fine in Windows (64 bit VS 2015, both Release and Debug). I'm seeing:
- boost::dynamic_bitset in broadphase/morton.h

And in test:
- boost::unit_test
- boost::timer
- boost::filesystem
- boost::dynamic_bitset

And nothing else!

Ready to merge this PR.

src/profile.cpp
```diff
@@ -222,7 +222,7 @@ void fcl::tools::Profiler::printThreadInfo(std::ostream &out, const PerThread &d
   {
     const AvgInfo &a = data.avg.find(avg[i].name)->second;
     out << avg[i].name << ": " << avg[i].value << " (stddev = " <<
-      sqrt(fabs(a.totalSqr - (double)a.parts * avg[i].value * avg[i].value) / ((double)a.parts - 1.)) << ")" << std::endl;
+      sqrt(std::fabs(a.totalSqr - (double)a.parts * avg[i].value * avg[i].value) / ((double)a.parts - 1.)) << ")" << std::endl;
```
sherm1: This should be `std::abs` rather than `std::fabs`.

Issue #107, opened 2016-03-26
test_fcl_math fails on Windows 64 bit due to non-portable use of long


sherm1:
Visual C++ defines `long` to be 32 bits even in 64 bit mode while gcc and clang make it 64 bits. The standard permits this and as a result the `long` type is generally useless and causes portability problems.
`long long` is 64 bits on all of these compilers. Generally we can avoid trouble by using the C++11 `<cstdint>` types like `uint64_t` when we're depending on a particular number of bits.

The `test_fcl_math` regression test fails on Win64 because it assumes [here](https://github.com/flexible-collision-library/fcl/blob/master/test/test_fcl_math.cpp#L639) that a 60-bit bitset can be returned in an `unsigned long`. `boost::dynamic_bitset` doesn't support `long long` but `std::bitset` does.

I'm working on removing dynamic_bitset and will fix this issue with that PR.

PR #108, opened 2016-03-26
Removes dependence on boost::dynamic_bitset and fixes test failure on Windows

sherm1:
This PR changes the bitset specialization of `morton_functor` to use `std::bitset` instead of `boost::dynamic_bitset`. The number of bits must now be set at compile time but there doesn't seem to be any reason for run time sizing, and it wasn't being used anywhere or tested.

When this is merged there will no longer be any boost dependency in the fcl code. However, the tests are still using the boost test framework.

This PR also fixes a Windows test failure reported in #107. That reduces the number of Windows failures from 4 to 3.

### DIFF
--- a/include/fcl/broadphase/morton.h
+++ b/include/fcl/broadphase/morton.h
@@ -156,7 +156,7 @@ struct morton_functor<FCL_UINT64>
 
 /// @brief Functor to compute N bit morton code for a given AABB
 /// N must be a multiple of 3.
-template<int N>
+template<size_t N>
 struct morton_functor<std::bitset<N>> 
 {
   static_assert(N%3==0, "Number of bits must be a multiple of 3");
--- a/include/fcl/broadphase/morton.h
+++ b/include/fcl/broadphase/morton.h
@@ -3,6 +3,7 @@
  *
  *  Copyright (c) 2011-2014, Willow Garage, Inc.
  *  Copyright (c) 2014-2015, Open Source Robotics Foundation
+ *  Copyright (c) 2016, Toyota Research Institute
  *  All rights reserved.
  *
  *  Redistribution and use in source and binary forms, with or without
@@ -38,9 +39,10 @@
 #ifndef FCL_MORTON_H
 #define FCL_MORTON_H
 
-#include <boost/dynamic_bitset.hpp>
 #include "fcl/data_types.h"
 #include "fcl/BV/AABB.h"
+
+#include <bitset>
 
 namespace fcl
 {
@@ -93,6 +95,9 @@ static inline FCL_UINT64 morton_code60(FCL_UINT32 x, FCL_UINT32 y, FCL_UINT32 z)
 
 
 /// @brief Functor to compute the morton code for a given AABB
+/// This is specialized for 32- and 64-bit unsigned integers giving
+/// a 30- or 60-bit code, respectively, and for `std::bitset<N>` where
+/// N is the length of the code and must be a multiple of 3.
 template<typename T>
 struct morton_functor {};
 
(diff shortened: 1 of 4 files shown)


include/fcl/broadphase/morton.h
```diff
@@ -145,50 +150,52 @@ struct morton_functor<FCL_UINT64>
...
+  static constexpr size_t bits() { return 60; }
 };
 
-/// @brief Functor to compute n bit morton code for a given AABB
-template<>
-struct morton_functor<boost::dynamic_bitset<> >
+
+/// @brief Functor to compute N bit morton code for a given AABB
+/// N must be a multiple of 3.
+template<int N>
```
mamoll: This is rather nitpicky, but maybe change to `template<std::size_t N>`?

mamoll:
Looks good to me. Any idea why the gcc builds fail on Travis? Is gcc 4.8 perhaps too old?

include/fcl/broadphase/morton.h
sherm1: Not nitpicky -- you are right! It should have had the same type as the template argument for `std::bitset`. Will fix.

sherm1:
> Any idea why the gcc builds fail on Travis?

Looks like it didn't match the bitset template argument to the corresponding specialization. I'll investigate.

sherm1:
@mamoll -- I fixed the template argument. While investigating the compile problem gcc 5 gave me some helpful warnings that uncovered a problem with the new definitions for the fixed-size int types (see commit message). With those two fixes, the build succeeds on gcc 4.8, clang 3.5, and VC++ 2015 (although there are still 3 test failures on Windows).

Please re-review.

include/fcl/data_types.h
```diff
@@ -45,10 +45,10 @@ namespace fcl
...
 
 typedef double FCL_REAL;
-typedef std::uint_fast64_t FCL_INT64;
-typedef std::int_fast64_t FCL_UINT64;
-typedef std::uint_fast32_t FCL_UINT32;
-typedef std::int_fast32_t FCL_INT32;
+typedef std::int64_t  FCL_INT64;
+typedef std::uint64_t FCL_UINT64;
+typedef std::int32_t  FCL_INT32;
+typedef std::uint32_t FCL_UINT32;
```
mamoll: Good thing you caught the backwards definition! That was my bad. According to http://en.cppreference.com/w/cpp/types/integer std::int64_t et al. are an optional part of the C++11 and are thus not guaranteed to be defined, whereas the \*fast\* versions are. What problems occur if there are too many bits?

sherm1: > std::int64_t et al. are an optional part of the C++11 standard ...

They are only optional on platforms that can't provide an exactly 32 & 64 bit 2's complement type. According to C++11 7.20.1.1/3 (quoting from stackoverflow):
> However, if an implementation provides integer types with widths of 8, 16, 32, or 64 bits, no padding bits, and (for the signed types) that have a two’s complement representation, it **shall define** the corresponding typedef names.

A more general solution would be to use the "least" types like `std::int_least32_t` which will be exactly the specified if it is available, otherwise the next larger size. The problem with that IMO is that to me the 32 in `FCL_INT32` implies exactly 32 bits by analogy with `int32_t`. I would rename it `FCL_INT_LEAST32` if that was what it means. Actually I would prefer just to remove these macros altogether and use the standard typedefs instead (in another PR).

> What problems occur if there are too many bits?

That would depend on the usage. I believe in the Morton case, the 32 & 64 bit specializations were intended to provide compact storage, so choosing a 64 bit representation for the 32 would be odd (even if that were the fastest type). I would just like the integer size dependencies to be clear. Currently this is kind of a mess in FCL and I'm wondering whether that may account for some of the failures I'm seeing on Windows.

Here are some related discussions: [Google style guide](https://google.github.io/styleguide/cppguide.html#Integer_Types), [Khronos issue](https://github.com/KhronosGroup/Vulkan-Docs/issues/122)

mamoll: What a mess indeed. I am satisfied that your suggested change is the best (or least bad) choice.

PR #110, opened 2016-03-28
Fixes a nasty bug in propagateBVHFrontListCollisionRecurse()

sherm1:
This was causing `test_fcl_frontlist` to seg fault on Windows. The bug is platform independent though so it definitely wasn't working right anywhere. Windows happened to access unreadable memory; probably it picked up garbage elsewhere. Apparently the test case isn't too fussy!

Also did some minor code cleanup (std::fabs -> std::abs, fix a typo).

@panjia1983, do you know how the test case could be improved so that it would have noticed this?

### DIFF
--- a/src/ccd/motion.cpp
+++ b/src/ccd/motion.cpp
@@ -54,12 +54,12 @@ FCL_REAL TBVMotionBoundVisitor<RSS>::visit(const SplineMotion& motion) const
 
   FCL_REAL tmp;
   // max_i |c_i * n|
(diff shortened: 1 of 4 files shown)


sherm1:
With this merged there are only two remaining test failures on windows:
- test_fcl_geometric_shapes has a tolerance issue (changing a 1e-5 to 1e-4 makes it pass)
- test_fcl_broadphase fails (no diagnosis yet)

mamoll:
:+1:

Issue #96, opened 2016-03-16
Using the Eigen library for math operations


jmirabel:
Hello,

I have experimented integrating the [Eigen](http://eigen.tuxfamily.org) library and I'm wondering whether you are interested in the resulting code.
Below are more details.

# Motivation
The greatest advantage in using Eigen in FCL is that it avoids creating non-necessary temporary matrices (and vectors). The following code:
```
Vec3f a, b;
Vec3f c = a + b;
```
is, with FCL vectors, compiled as:
```
tmp <- a.operator+(b);
c <- tmp
```
whereas, with Eigen vectors:
```
c <- a.operator+(b)
```
This becomes more interesting on more complex expressions. See [this](http://eigen.tuxfamily.org/dox/TopicInsideEigenExample.html) for more details on temporary elimination and lazy assignment.

# Preliminary results
So far, I have only integrated Eigen in `Vec3f` and `Matrix3f` and the improvements are mostly marginal, from 1% to 10% on the basic tests I did. The advantages of the method are:
- No copy required for users needing Eigen in a library which depends on FCL.
- A greater gain can be expected by relying on Eigen for other math classes such as `Quaternion3f`... I did not do any profiling to check what could be done.

# Implementation details
There are very few impacts in the current code. Three parts of the code needs to be changed, whenever the prepocessor variable `FCL_HAVE_EIGEN` is 1:
- Classes `Vec3f` and `Matrix3f` (so far) are declared based on Eigen, with the exact same interface. (the main problem is the inplace transpose in FCL)
- Class `TVector3` needs one additional `operator*`
- Conditional operator `?:;` does not work with expressions of matrices. For instance,
`b = (condition)?a:-a;` must be changed in `if (condition) b = a; else b = -a;`

jmirabel:
It seems there is some interest in this. I'll wait to see the conclusion of the discussion #97. My experiments were done on top the [humanoid-path-planner/hpp-fcl](https://github.com/humanoid-path-planner/hpp-fcl) fork, mentioned in #97.
Thanks

traversaro:
I have a question related to this as a (future) user of FCL in a project that is also using Eigen. 

[...]

PR #111, opened 2016-03-31
Optionnally use Eigen for math operations

jmirabel:
This refers to the issue #96.

List of classes that can use Eigen:
- [x] `Vec3f`, `Matrix3f`
- [ ] `Quaternion3f`: I haven't checked whether it is possible. A possible issue is that Eigen stores quaternions as `(x, y, z, w)` and FCL stores `(w, x, y, z)`.
- [ ] `Transform3f`
- [ ] `TaylorModel`, `TVector3` and `TMatrix3`


jmirabel:
I forgot to add: the CMake will very likely fail to find Eigen3 on windows.
I'm sure it's not hard to fix but I have no way of testing it.
Could someone have a look at it ?

sherm1:
I'll try it on Windows.

jslee02:
It seems there are no noticeable performance differences yet [with](https://travis-ci.org/flexible-collision-library/fcl/jobs/119866407#L1340) and [without](https://travis-ci.org/flexible-collision-library/fcl/jobs/119866406#L1340) Eigen according to the test running times in release mode. We might want to [take advantage of SIMD support of Eigen](http://eigen.tuxfamily.org/index.php?title=FAQ#How_can_I_enable_vectorization.3F) for better performance.

Also, using Eigen in debug mode is [much slower](https://travis-ci.org/flexible-collision-library/fcl/jobs/119866415#L1339) so that the travis test keeps failing. It seems travis regards a test failed when there is no output for 10 min. One possible workaround would be splitting the big tests into smaller ones?

mamoll:
Re. SIMD support, this can be done in CMake via [CheckCXXCompilerFlag](https://cmake.org/cmake/help/v3.5/module/CheckCXXCompilerFlag.html). I bet some other project already has some cmake logic to set the appropriate SSE/Altivec type flags. Maybe `-march=native` is enough (which can be enabled with -DFCL_USE_SSE=ON)?

scpeters:
We have some SSE detection code in gazebo, but I'm not sure how well it works:

* https://bitbucket.org/osrf/gazebo/src/default/cmake/FindSSE.cmake
* https://bitbucket.org/osrf/gazebo/src/default/cmake/HostCFlags.cmake

sherm1:
> Also, using Eigen in debug mode is much slower so that the travis test keeps failing

@jslee02, it looks like that is only happening for the coveralls build? The rest of the runs [here](https://travis-ci.org/flexible-collision-library/fcl/builds/119866404) seemed reasonable and lots of them were also Debug.

jslee02:
I would like to see how's the combination of debug + Eigen without coveralls option. I'll test it on linux and post the result here.

jslee02:
It seems debug + Eigen is slower regardless of coveralls option. Here is the results:

debug + without Eigen
```
Running tests...
Test project /home/js/dev/pr/fcl/eigen/build/Debug_without_eigen
      Start  1: test_fcl_collision
 1/14 Test  #1: test_fcl_collision ................   Passed   25.31 sec
      Start  2: test_fcl_distance
 2/14 Test  #2: test_fcl_distance .................   Passed   21.02 sec
      Start  3: test_fcl_geometric_shapes
 3/14 Test  #3: test_fcl_geometric_shapes .........   Passed    0.06 sec
      Start  4: test_fcl_broadphase
 4/14 Test  #4: test_fcl_broadphase ...............   Passed  117.92 sec
      Start  5: test_fcl_shape_mesh_consistency
 5/14 Test  #5: test_fcl_shape_mesh_consistency ...   Passed    1.12 sec
      Start  6: test_fcl_frontlist
 6/14 Test  #6: test_fcl_frontlist ................   Passed   17.56 sec
      Start  7: test_fcl_math
 7/14 Test  #7: test_fcl_math .....................   Passed    0.01 sec
      Start  8: test_fcl_sphere_capsule
 8/14 Test  #8: test_fcl_sphere_capsule ...........   Passed    0.01 sec
      Start  9: test_fcl_capsule_capsule
 9/14 Test  #9: test_fcl_capsule_capsule ..........   Passed    0.01 sec
      Start 10: test_fcl_simple
10/14 Test #10: test_fcl_simple ...................   Passed    0.01 sec
      Start 11: test_fcl_capsule_box_1
11/14 Test #11: test_fcl_capsule_box_1 ............   Passed    0.01 sec
      Start 12: test_fcl_capsule_box_2
12/14 Test #12: test_fcl_capsule_box_2 ............   Passed    0.01 sec
      Start 13: test_fcl_bvh_models
13/14 Test #13: test_fcl_bvh_models ...............   Passed    0.01 sec
      Start 14: test_fcl_octomap
14/14 Test #14: test_fcl_octomap ..................   Passed  178.79 sec

100% tests passed, 0 tests failed out of 14

Total Test time (real) = 361.83 sec
```

debug + with Eigen
```
Running tests...
Test project /home/js/dev/pr/fcl/eigen/build/Debug
      Start  1: test_fcl_collision
 1/14 Test  #1: test_fcl_collision ................   Passed  237.39 sec
      Start  2: test_fcl_distance
// still running...
```

sherm1:
> Total Test time (real) = 2425.85 sec

Ouch! Although a 10X Debug slowdown isn't uncommon for code that requires extensive inlining for performance.

include/fcl/math/matrix_3fx.h
```diff
@@ -0,0 +1,462 @@
...
+
+#ifndef FCL_MATRIX_3FX_H
+#define FCL_MATRIX_3FX_H
+
+#include "fcl/math/vec_3f.h"
+
+namespace fcl
+{
+
+/// @brief Matrix2 class wrapper. the core data is in the template parameter class
```
sherm1: Matrix3?

sherm1:
I'm confused about the size of this PR. It is hard to tell what's new and what got moved. Also there seemed to be a lot of duplicate code, such as an explicit eigenvalue computation.

I'm also wondering if it is necessary to inject all that Fcl code into Eigen's headers. The result is that FCL appears still to have its own matrix types rather than using Eigen's. In my mind that defeats the purpose -- I didn't expect there to be much of a speedup but it would be great if Eigen-using calling code didn't have to translate to and from Eigen objects to use Fcl.

jmirabel:
I see I have to express the motivations and expectations of my work a bit :)

# Motivations
I'm developing a library that depends on FCL and uses Eigen for math operations. The first thing I tried doing in FCL is activating the `FCL_USE_SSE` in CMake. That turned out to slow down FCL.

Vectorization for Vec3f and Matrix3f decreases the number of flops of some operations by 33% but increases the allocated memory by 25%. I suspect that cache misses were more important than the flops count on my processors. It might be different on a processor with bigger cache.
A link to the Eigen documentation was provided earlier. Eigen [does not vectorize](http://eigen.tuxfamily.org/dox/group__TopicFixedSizeVectorizable.html) fixed size matrices 3D vectors and 3x3 matrices.

As I stated in #96, the main interest in using Eigen is the lazy assignment and the temporary elimination (when relevant).

# Design choices
The main issue was that FCL and Eigen uses different convention for `transpose` and for `inverse`.
FCL is doing both inplace, while Eigen is not. So there was 2 options:
- change in all the code of FCL the `transpose` and `inverse` calls and add Eigen as a required dependency. This choice is yours, not mine.
- Inherits from `Eigen::Matrix` or `Eigen::MatrixBase`. I found the first was easier and generated a bit less code.
As @sherm1 said, there is some duplicated code. This is mostly because I intended to "Eigenize" it, but did not have time to do it. Sorry about that.

# Eigen and debug mode
Yes, `Eigen` in debug mode is dramatic (both compilation time and runtime) ! That is not surprising, the number of implementation of template classes is quickly exploding.

# Performances expectations
Using unit tests to compare performances is not satisfactory. I haven't try to figure out the portion of time spend in initialization or in other libraries. When I claimed a gain between 1 and 10%, I was not talking about the unit tests, I was using [another library](https://github.com/stack-of-tasks/pinocchio/) which among other providing a benchmark of, among other, collision and distance computation in FCL (see [this file](https://github.com/stack-of-tasks/pinocchio/blob/devel/benchmark/timings-geometry.cpp) ).
It is a pretty basic test, but, IMHO, it is more reliable than unit tests.

Finally, the performances of Eigen cannot be fully exploited without a bit of code translation. Among other, changing `Vec3f a = b` in `Vec3f a (b)`... Eigen documentation provides some tips (like http://eigen.tuxfamily.org/dox/TopicWritingEfficientProductExpression.html)

include/fcl/math/matrix_3fx.h
jmirabel: Oops, yes Matrix3. That is a copy paste of [this](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/math/matrix_3f.h#L46)

jmirabel:
I discovered that there was some timings done in the unit tests. To ease the discussion, I added a commit with a benchmark output of the 3 configurations (FCL, Eigen, FCL+SSE) and of the script I used to generate them. I will remove it on request.

With the compilation option `-march=native`, all three configurations fails at `test_fcl_geometric_shapes` and the Eigen configuration is doing marginally better with  `-march=native` than without.

Comparing `benchmark.build-eigen.log` and `benchmark.build-orig.log`, there seems to be a marginal advantage for Eigen. But I am not sure of how these outputs should be compared.

sherm1:
@jmirabel, thanks for the clarification -- much appreciated!

I think performance is the wrong goal here. It seems unlikely to me that conversion to Eigen will help much although it could maybe after some Eigen-specific tweaking of the code. Instead, I think the goal should be to replace Fcl's home-made matrix code with the commonly used Eigen equivalents. The result *should* be: (1) substantial shrinking of Fcl's code base; (2) familiar and predictable behavior for methods like `transpose()`; (3) access to Eigen's functionality and performance features, including those being actively developed like AVX support (which is much more effective than SSE).

> change in all the code of FCL the transpose and inverse calls and add Eigen as a required dependency. This choice is yours, not mine.

I think these are dangerous methods as currently implemented. I would propose as a first step a PR that renames all of them to something like `transposeInPlace()` and `invertInPlace()`. That way there will be no confusion about what they are doing, and they won't get mixed up with Eigen methods of the same name.

In general I think this project requires a series of small PRs rather than one big one, with the goal first to clean up the Fcl code to pave the way for the Eigen conversion, then introduce an unused Eigen3 dependency to the build, then convert in small increments (if possible) until the Fcl custom-math stuff can be removed.

What do you think?

mamoll:
I haven't looked at the PR carefully, but I agree with @sherm1. It looks like the PR adds Eigen support but keeps some version of the current code if Eigen isn't detected at build time. I'd favor an unconditional dependency on Eigen and reduce the code base.

jmirabel:
If there is no special problem with adding a required dependency, I totally agree with removing the current FCL mathematical classes for Eigen ones. Less code is better. This PR does not go into this direction.

PS: I think you are under-estimating the possible gain of carefully implementing mathematical operations.

sherm1:
> PS: I think you are under-estimating the possible gain of carefully implementing mathematical operations.

Good! I hope you're right.

PR #114, opened 2016-04-04
[BUG FIX]octree vs mesh CollisionResult now returns triangle id.

stonneau:
In previous version of fcl, the Contact objects contained in CollisionResult object, pointed
to invalid triangle id in a specific case.
This case is the collision request between an octree and a mesh.
This has been fixed and a simple test has been added.
It simply verifies that the returned triangle id actually exists in the object.


sherm1:
:+1: Thanks, @stonneau. Merging since no one has objected.

Issue #119, opened 2016-04-05
Motion initialization and interpolation


schmrlng:
https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/ccd/motion.h defines some classes for interpolating between ```Transform3f```s that are the basis of performing continuous collision checking. Currently, if you call ```getCurrentTransform``` for a TranslationMotion before ```integrate``` has been called, you get an identity transformation which causes the CCDC_CONSERVATIVE_ADVANCEMENT mode of ```continuousCollide``` to always return true ([here](https://github.com/flexible-collision-library/fcl/blob/master/src/ccd/conservative_advancement.cpp#L231), ```getCurrentTransform``` is called before ```integrate```).

I'll make a pull request addressing that, but I also see a discrepancy with ```ScrewMotion``` interpolation (the mode used by [omplapp](https://github.com/ompl/omplapp/blob/master/src/omplapp/geometry/detail/FCLMethodWrapper.h#L117)) compared to ```TranslationMotion``` or ```InterpMotion```:

> -------------- InterpMotion --------------
getCurrentTransform at motion initialization
T = (1 1 1)
R = [(1 0 0)(0 1 0)(0 0 1)]
q = (1 0 0 0)
InterpMotion interpolated at t = 0
T = (1 1 1)
R = [(1 0 0)(0 1 0)(0 0 1)]
q = (1 0 0 0)
InterpMotion interpolated at t = 0.2
T = (1.2 1.2 1.2)
R = [(1 0 0)(0 1 0)(0 0 1)]
q = (1 0 0 0)
InterpMotion interpolated at t = 0.4
T = (1.4 1.4 1.4)
R = [(1 0 0)(0 1 0)(0 0 1)]
q = (1 0 0 0)
InterpMotion interpolated at t = 0.6
T = (1.6 1.6 1.6)
R = [(1 0 0)(0 1 0)(0 0 1)]
q = (1 0 0 0)
InterpMotion interpolated at t = 0.8
T = (1.8 1.8 1.8)
R = [(1 0 0)(0 1 0)(0 0 1)]
q = (1 0 0 0)
InterpMotion interpolated at t = 1
T = (2 2 2)
R = [(1 0 0)(0 1 0)(0 0 1)]
q = (1 0 0 0)
-------------- ScrewMotion --------------
getCurrentTransform at motion initialization
T = (1 1 1)
R = [(1 0 0)(0 1 0)(0 0 1)]
q = (1 0 0 0)
[...]

PR #121, opened 2016-04-05
Initialize tf in TranslationMotion

schmrlng:
This is consistent with the implementations of ```InterpMotion``` and ```ScrewMotion```.


schmrlng:
This also addresses the first portion of https://github.com/flexible-collision-library/fcl/issues/119.

PR #125, opened 2016-04-22
Hide coveralls repository token

jslee02:
The coveralls repository token is supposed to be private because it grants write access. To hide it while it keeps working:

1. removed `COVERALLS_REPO_TOKEN` from `.travis.yml`
1. generated a new repository token
1. set `COVERALLS_REPO_TOKEN` to the new token on Travis as a hidden environment variable

as [we did in DART](https://github.com/dartsim/dart/pull/688) thanks to @mkoval.

Also, I changed the coverage test to be done in linux/debug builds instead of adding a separate build to reduce the whole Travis test. It could be better the coverage test is done only for gcc, but Travis doesn't provide environment variable to detect the compiler. I'm open to hearing a better way to do this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,14 +14,10 @@ compiler:
   - clang
 
 env:
-  - BUILD_TYPE=Debug   COVERALLS=OFF
+  - BUILD_TYPE=Debug   COVERALLS=ON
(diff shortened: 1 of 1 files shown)


sherm1:
Looks good to me.

PR #135, opened 2016-06-24
add option to disable octomap in build

v4hn:
This is a quasi-standard pattern to allow users to test the library without
its optional-dependency even if the dependency is available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,32 +66,38 @@ endif()
 
 # Find Octomap (optional)
 find_package(PkgConfig QUIET)
+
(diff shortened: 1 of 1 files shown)


jslee02:
This looks good to me.

nitpick: Could we add `FCL_` prefix for the option name for consistency?

v4hn:
done

jslee02:
another nitpick: For the default value, I believe `TRUE` works, but `ON` seems [recommended one](https://cmake.org/cmake/help/v3.0/command/option.html) and better for consistency.

v4hn:
Hehe, there you go, oh great maintainer of this wonderful library. :-)

Issue #132, opened 2016-06-20
error: missing binary operator before token "("


dblanm:
When making make in the build directory, using sudo make, make or make -j4
It gives me the error 
home.../fcl/octree.h error: missing binary operator before token "("
 #if OCTOMAP_VERSION_AT_LEAST(1,8,0)
                                                            ^

.... and it follows 

I change it to  #if OCTOMAP_VERSION_AT_LEAST(1.8.0)
but it didn't worked
Anyone with same problem that knows how to fix it?

Thanks in advance!

dblanm:
The code failing is this one:

 /// @return ptr to child number childIdx of node
  OcTreeNode* getNodeChild(OcTreeNode* node, unsigned int childIdx)
  { 
#if OCTOMAP_VERSION_AT_LEAST(1,8,0)
    return tree->getNodeChild(node, childIdx);
#else
    return node->getChild(childIdx);
#endif
} 

I have to say that I have installed last version of octomap, and checked it in /usr/local/share/octomap/package.xml

jslee02:
Could you try a clean build [as we did](https://github.com/flexible-collision-library/fcl/pull/129#issuecomment-226519520)?

dblanm:
I am sorry but I am a bit new modifying this.
I modified the fcl/config.h.in and added:
#if FCL_HAVE_OCTOMAP
  #define OCTOMAP_MAJOR_VERSION @OCTOMAP_MAJOR_VERSION@
  #define OCTOMAP_MINOR_VERSION @OCTOMAP_MINOR_VERSION@
  #define OCTOMAP_PATCH_VERSION @OCTOMAP_PATCH_VERSION@

  #define OCTOMAP_VERSION_AT_LEAST(x,y,z) \
    (OCTOMAP_MAJOR_VERSION > x || (OCTOMAP_MAJOR_VERSION >= x && \
    (OCTOMAP_MINOR_VERSION > y || (OCTOMAP_MINOR_VERSION >= y && \
    OCTOMAP_PATCH_VERSION >= z))))

  #define OCTOMAP_VERSION_AT_MOST(x,y,z) \
    (OCTOMAP_MAJOR_VERSION < x || (OCTOMAP_MAJOR_VERSION <= x && \
    (OCTOMAP_MINOR_VERSION < y || (OCTOMAP_MINOR_VERSION <= y && \
    OCTOMAP_PATCH_VERSION <= z))))
#endif // FCL_HAVE_OCTOMAP

[...]

PR #141, opened 2016-07-21
Backporting #53 to FCL 0.3

jslee02:
as [requested](https://github.com/flexible-collision-library/fcl/pull/53#issuecomment-234392430).

What I'm not sure is how `FCL_ABI_VERSION` should be updated.

### DIFF
--- a/CMakeModules/FCLVersion.cmake
+++ b/CMakeModules/FCLVersion.cmake
@@ -1,7 +1,7 @@
 # set the version in a way CMake can use
 set(FCL_MAJOR_VERSION 0)
 set(FCL_MINOR_VERSION 3)
-set(FCL_PATCH_VERSION 2)
+set(FCL_PATCH_VERSION 3)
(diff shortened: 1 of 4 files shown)


wjwwood:
@jslee02 I was just trying this locally and I ran into this issue: https://github.com/flexible-collision-library/fcl/issues/132

I looked through our release pipeline and I don't see any patches to address this... It's strange to me that what we're using to release 0.3.2 is compiling, but this branch is not compiling locally...

wjwwood:
Actually, the `0.3.2` tag doesn't compile either :fearful:. I'll keep looking into this issue.

wjwwood:
I figured out my issue. This is a problem anytime you have fcl installed to the same place as one of the dependencies, like octomap, and the installed `config.h` is different from the one you generated while building. I'll open a pr.

jslee02:
Interesting. I've never encountered the issue even though I think I had fcl installed when building another fcl. Anyways, you resolved the mysterious building issue (#132) to me. Thanks!

PR #142, opened 2016-07-21
ensure the locally generated config.h is used

wjwwood:
Should fix https://github.com/flexible-collision-library/fcl/issues/132

This is necessary, because if one of the dependencies is installed to a place where another version of `fcl` is also installed, it can happen that the include directories favor the installed `fcl/config.h` over the one generated for the local build. This is why people never saw it in CI (since CI didn't install `fcl` before trying to build `fcl`).

This should be backported to the `fcl-0.3` branch too :smile:.


jslee02:
:+1:

PR #143, opened 2016-07-22
Use major.minor version for ABI soversion

j-rivero:
There is currently a custom version associated with ABI bumps. While it is not an incorrect approach it can be improved by using a common schema usually associated with the software version. According to historic development it looks to me that FCL follow something like major.minor.micro versioning scheme with ABI guarantee during major.minor.

I've been patching the sources in the debian package to generate SONAME according to major.minor so I thought that could be a good idea to submit this PR for discussion and adoption if we agree in a versioning policy.


jslee02:
:+1:

PR #146, opened 2016-07-23
Remove Boost Test and replace with Google Test

jslee02:


sherm1:
This looks great, JS -- I'm really happy to see boost evaporating! One suggestion: rather than using Google's `TEST` macro use `GTEST_TEST`. `TEST` is a mistake Google made (in violation of its own standards I think) in creating a macro with such a common name that it conflicts with other code. Hence the need for [this kludge](https://github.com/google/googletest/blob/master/googletest/docs/FAQ.md#google-test-defines-a-macro-that-clashes-with-one-defined-by-another-library-how-do-i-deal-with-that) in the google test manual. Other mistakes were `SUCCEED` and `FAIL`. I suggest defining the provided `GTEST_DONT_DEFINE_TEST` etc. macros to avoid the potential for clash with third party or user code.

jslee02:
That's a good suggestion! Done.

PR #153, opened 2016-08-08
Fix incorrect use of FCL_BUILD_TYPE_DEBUG

jslee02:
The build type check with `FCL_BUILD_TYPE_DEBUG` and `FCL_BUILD_TYPE_RELEASE` should be done using their values instead of whether they are defined or not since they are always defined.


sherm1:
Change looks good.

Question: is it useful to have these macros in addition to the language-standard `NDEBUG` macro?

jslee02:
> Question: is it useful to have these macros in addition to the language-standard NDEBUG macro?

When I added those macros, I considered the case FCL would like to have more build types (e.g., `BUILD_TYPE_RELWITHDEBINFO` and `BUILD_TYPE_MINSIZEREL`) with different behaviors, but haven't found any use case so far. I'm not against removing them.

sherm1:
Could be a topic for a different PR.

PR #154, opened 2016-08-09
Templatize FCL for scalar type

jslee02:
This PR templatize FCL objects on the scalar type so that we can take advantage of Eigen's [automatic differentiation module](https://eigen.tuxfamily.org/dox/unsupported/group__AutoDiff__Module.html) as [suggested](https://github.com/flexible-collision-library/fcl/issues/96#issuecomment-234699398) by @sherm1 . A simple unit test that computes the gradient of distance w.r.t. the object position using Eigen's AutoDiff Module is included.

To make FCL working fine with `Eigen::AutoDiffScalar`, I had to templatize most of the FCL classes and global functions. Now there are only 7 files not templatized over ~180 files, which are not used anyways. I would suggest making FCL header only project as future change.

In this work, the major challenge was the resolving circular dependencies as all the implementations of templated classes should be placed in the same file with the declarations. To resolve the problem, I split those header files in circular dependencies into several files. Also, some files are split for easier maintenance (e.g., geometric_shapes.h is split per each shape class).

One caveat is the increased compilation time, which is the nature of header only project.


sherm1:
Wow! This is awesome, JS.

sherm1:
FYI (just for your consideration): we had a long discussion about these template arguments in [this Drake issue](https://github.com/RobotLocomotion/drake/issues/2255). We concluded that `Scalar` is awkward due to `AutoDiffScalar::Scalar` and settled on just plain old `T` which has worked out very nicely. It has the effect of demoting the template argument so that it doesn't dominate so much space (physical and mental). And we can get the actual scalar type with `T::Scalar`.

jslee02:
Thanks! As this change is huge, there might be some oversights. Any feedback would be very helpful.

Regarding template argument: I quickly went through the discussion and have the same concern on using `T` in the sense of that "`T` implies that any type is acceptable, which isn't true here". For the template argument, I thought we only expect primitive floating point types or any object behaves like floating point type (e.g., `Eigen::AutoDiffScalar`) so that we don't need to demote the template argument from `Scalar`. Please let me know if I'm missing something.

sherm1:
> T implies that any type is acceptable, which isn't true here. 
> Please let me know if I'm missing something.

Well, you can see how the argument went on from there in [Drake 2255](https://github.com/RobotLocomotion/drake/issues/2255). `Scalar` at first seemed like the obvious thing but suffers from the Eigen-induced ambiguity that all Eigen "scalar" types must have a typename member called `Scalar`. So there is `AutoDiffScalar<DerType>::Scalar`. If we call the template argument Scalar then we get `Scalar::Scalar` which is, well, weird. 

The other winning argument had to do with visual code bloat. When every class is templatized by the numeric type, it should fade into the background since it is not important. We felt that using a long name ended up obscuring a lot of the more important code because there would be several `Scalar`s on some lines of code, using a significant fraction of the 80 characters we are allocated. Basically, the code ends up being much easier to read when the (trivial) numeric type doesn't occupy much space.

Finally, it is very common for type `T` to be used in template classes and to restrict that type to those that satisfy particular "concepts", ideally enforced with `static_assert` when possible. In common usage it doesn't really imply that *any* type is allowed, just that the template argument is some type.

But ... I think it is fine if you prefer `Scalar`. It has some drawbacks but you are well aware of them. It also has some good features, so it's just a judgment call.

sherm1:
Building on Windows VS 2015 64 bit, Debug using Eigen 3.2.93 I got one build failure in `test_fcl_auto_diff.cpp`, beginning:
```
c:\eigen3.2.93\eigen\src/Geometry/Transform.h(1371): error C2666: 'Eigen::AutoDiffScalar<derivative_t>::operator *': 3 overloads have similar conversions
6>  c:\eigen3.2.93\unsupported\eigen\src/AutoDiff/AutoDiffScalar.h(274): note: could be 'const Eigen::AutoDiffScalar<Eigen::CwiseBinaryOp<Eigen::internal::scalar_product_op<float,float>,const Eigen::Matrix<float,3,1,0,3,1>,const Eigen::CwiseNullaryOp<Eigen::internal::scalar_constant_op<float>,const Eigen::Matrix<float,3,1,0,3,1>>>> Eigen::AutoDiffScalar<derivative_t>::operator *(const float &,const Eigen::AutoDiffScalar<derivative_t> &)' [found using argument-dependent lookup]
6>  c:\eigen3.2.93\eigen\src/Core/GeneralProduct.h(369): note: or       'const Eigen::Product<Derived,Eigen::Matrix<Eigen::AutoDiffScalar<derivative_t>,4,1,0,4,1>,0> Eigen::MatrixBase<Derived>::operator *<Eigen::Matrix<Eigen::AutoDiffScalar<derivative_t>,4,1,0,4,1>>(const Eigen::MatrixBase<Eigen::Matrix<Eigen::AutoDiffScalar<derivative_t>,4,1,0,4,1>> &) const'
6>          with
6>          [
6>              Derived=Eigen::Matrix<Eigen::AutoDiffScalar<derivative_t>,3,4,0,3,4>
6>          ]
```
and then going on and on as is common with Eigen.

I haven't investigated yet -- did you possibly use a different version of Eigen?

jslee02:
I agree on the winning argument, the visual code bloat. FCL already suffers from other long-named templates arguments such as `NarrowPhaseSolver`. I believe the original author preferred informative name rather than short one. This is one reason I picked `Scalar` rather than a single letter like `T` or `S`. I don't have a strong preference on this. I would vote on `S` for single letter though.

I apologize for that I still don't see why we need to call the template argument `Scalar` from an object passed in as the scalar template argument. I think we use the object passed in treating it is a floating point variable (it could be `float`, `double`, and any object that behaves like floating point type) rather than assuming it is an `Eigen` object. So we don't need to get the scalar type from the template argument because the template argument itself should work as a scalar type. I would imagine the template argument demoting `Scalar::Scalar` would be failed if we pass primitive types such as `float` or `double`.

jslee02:
> did you possibly use a different version of Eigen?

I used Eigen 3.2.7. I'll take a look at it.

sherm1:
Re template arg naming: found a random FCL code example while chasing the compile error I mentioned:
```c++
template<typename Scalar>
struct ShapeDistanceLibccdImpl<Scalar, Sphere<Scalar>, Sphere<Scalar>> {};
// vs.
template<typename S>
struct ShapeDistanceLibccdImpl<S, Sphere<S>, Sphere<S>> {};
```
I think the meaning of the code comes through better that way.

jslee02:
Quick note: It works with Eigen 3.2.9 but not with Eigen 3.2.93, which is 3.3-beta2. Maybe it's not stable yet?

sherm1:
> I would imagine the template argument demoting Scalar::Scalar would be failed if we pass primitive types such as float or double.

Yes, you are right -- my bad. This seems to apply only to Eigen types, using the `internal::traits` helper class which is properly overloaded for all Eigen classes. AutoDiffScalar looks like this:
```c++
template<typename _DerType>
class AutoDiffScalar : // ...
    typedef typename internal::traits<DerType>::Scalar Scalar;
```
Then `Eigen::NumTraits<Scalar>::Real` gives the underlying type, even if `Scalar` is float or double.

jslee02:
One issue with using `typename S` for scalar type is that `typename S` and `typename T_SH` have been used for shape classes. I would prefer using `S` for scalar type and `Shape` for Shape classes. How do you think?

sherm1:
> Quick note: It works with Eigen 3.2.9 but not with Eigen 3.2.93, which is 3.3-beta2.

Thanks -- works now with 3.2.9!

sherm1:
Some performance comparisons on Win64 VS 2015 RelWithDebInfo:
I built FCL 0.5, FCL master (Eigen but not templatized), and this PR (Eigen plus templatized). I didn't measure compile times, but it did seem to get much slower with Eigen.

FCL 0.5/Eigen/Eigen+Templates:
```
Test  #1: test_fcl_collision ................   Passed    4.34/4.78/5.72
Test  #2: test_fcl_distance .................  Passed    3.78/4.33/4.73
Test  #4: test_fcl_broadphase .......  Passed  112.92/143.10/159.19 (but see below)
Test  #6: test_fcl_frontlist ................    Passed    2.94/3.43/4.17
```
The broadphase test is split in the templatized version, so possibly isn't comparable:
```
 test_fcl_broadphase_collision_1
2/16 Test  #2: test_fcl_broadphase_collision_1 ...   Passed   63.88 sec
3/16 Test  #3: test_fcl_broadphase_collision_2 ...   Passed   55.22 sec
4/16 Test  #4: test_fcl_broadphase_distance ......    Passed   40.09 sec
```

Not sure why this would be getting slower. Maybe there is lots more code generated now and we're getting instruction cache faults?

jslee02:
I optimized the code and the current performance of this PR seems comparable to FCL 0.5 now.

Here are what I get on:

A. OS X El Capitan + clang-703.0.31 + Release:

| | FCL 0.5 | FCL master (Eigen) | This PR (Eigen + templatizing) |
|---|---|---|---|
| test_fcl_collision | 4.33 | 4.26 | 4.21 |
| test_fcl_distance | 3.34 | 3.88 | 3.52 |
| test_fcl_broadphase | 120.6 | 129.11 | 122.9 (45.88+52.17+24.85) |
| test_fcl_frontlist | 2.93 | 2.99 | 2.87 |
| test_fcl_octomap | - | 75.01 | 56.07(4.39+18.38+33.3) |

B. Ubuntu Trusty + gcc 4.8.5 + Release:

| | FCL 0.5 | FCL master (Eigen) | This PR (Eigen + templatizing) |
|---|---|---|---|
| test_fcl_collision | 2.98 | 2.94 | 2.71 |
| test_fcl_distance | 2.48 | 2.71 | 2.24 |
| test_fcl_broadphase | 85.46 | 89.40 | 75.44 (29.15+32.45+13.84) |
| test_fcl_frontlist | 2.06 | 2.11 | 1.96 |
| test_fcl_octomap | 43.10 | 52.64 | 35.59 (2.62+11.48+21.49) |

Here is summary of the changes in the last commits:
* Applied [Eigen's noalias()](https://eigen.tuxfamily.org/dox/group__TopicAliasing.html) to where applicable and most frequently called code. There might be more spots the technique can be applied, which could be addressed in future PR.
* Replaced matrix multiplication of (3x3) matrix and (3x1) vector with three dot product (e.g., [this](https://github.com/flexible-collision-library/fcl/pull/154/commits/b9584cf0749e46b9bfd9b6928ef960160b9a8a79#diff-bf96101807f04b55cc69eb6795215494L130)). This is what I'm not sure why three dot products are faster than one matrix multiplication. Both of `v^T * axis`  and `axis.transpose() * v` showed the same performance, which is slower than three dot products.

One source of performance decrease might be [here](https://github.com/flexible-collision-library/fcl/blob/templatize/include/fcl/narrowphase/detail/gjk_libccd.h#L670). Since `libccd` only takes quaternions rather than rotation matrices, FCL has to convert rotation matrix to quaternion every time calls `libccd` routines, which is for convex-based shape collisions and distance queries. Previously, FCL stores quaternion and rotation matrix at the same time in `Transform3f`.

sherm1:
That looks great, JS. I'll try it again on Windows.

sherm1:
Win64 VS 2015 RelWithDebInfo
-----------------------------------
Re-run with your latest changes (had to fix a typo in spatial_hash.h). Results from above repeated; new ones marked with `**`.
```
                                             FCL 0.5/master/templatized/**optimized**
Test  #1: test_fcl_collision ........ Passed    4.34/4.78/5.72/**4.93**
Test  #2: test_fcl_distance ......... Passed    3.78/4.33/4.73/**3.84**
Test  #4: test_fcl_broadphase ....... Passed  112.92/143.10/159.19/**135.06** (breakdown below)
Test  #6: test_fcl_frontlist ........ Passed    2.94/3.43/4.17/**3.77**
```
The broadphase test is split in the templatized version, so possibly isn't comparable:
```
Test  #2: test_fcl_broadphase_collision_1 ...   Passed   63.88/**51.90**
Test  #3: test_fcl_broadphase_collision_2 ...   Passed   55.22/**57.87**
Test  #4: test_fcl_broadphase_distance ......    Passed   40.09/**25.29**
```

Not back to 0.5 levels, but generally much better.

jslee02:
Thanks for the comparisons! For fair comparisons, FCL 0.5 build should be updated with [this](https://github.com/flexible-collision-library/fcl/pull/155), but I don't it would make a big difference, though. Hm, I don't know why this PR is slower than 0.5 on Windows.. I could to some investigation on my virtual Windows machine, but I would like to save it for future work.

The typo is [fixed](https://github.com/flexible-collision-library/fcl/pull/156/commits/591582b78acc2e88a73ce4578dd4c58e59db24fd) in #156.

sherm1:
See issue #160 for a subsequent change we could make to lighten the compilation load created by templatizing.

jslee02:
The idea sounds good to me. I would like to save it for a future pull request, though.

Also, I believe this PR is ready to merge.

sherm1:
> The idea sounds good to me. I would like to save it for a future pull request, though.

Definitely! Could be done any time later.

> Also, I believe this PR is ready to merge.

I agree.

PR #157, opened 2016-08-13
Enable Win32 build on Appveyor

jslee02:


sherm1:
Great idea -- this should expose any Eigen alignment problems.

sherm1:
For maximally extreme testing, this should be done with a Debug build to enable Eigen's many assertions.

jslee02:
Yeah, Debug build is included for all the CI tests.

For Eigen alignment issue, I tested on Ubuntu Trusty 32bit + gcc + SSE2 enabled, and it seems to work fine. We might also need #159 to keep making sure it's being resolved.

PR #158, opened 2016-08-13
Replace NULL with nullptr

jslee02:
for type safety

### DIFF
--- a/include/fcl/BV/AABB.h
+++ b/include/fcl/BV/AABB.h
@@ -118,7 +118,7 @@ public:
   /// @brief Center of the AABB
   Vector3<S> center() const;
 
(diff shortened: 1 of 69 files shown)


sherm1:
:+1:

mamoll:
For more complex code transformations that take advantage of C++11 features, you should check out [clang-tidy](http://clang.llvm.org/extra/clang-tidy/). See my commits on July 28 in the [OMPL repo](https://github.com/ompl/ompl/commits/master) for examples.

jslee02:
Good idea! I will take a look at clang-tidy once #154 is merged since the code transformation would be applied all over the code base. I wasn't patient with `NULL` until that. :sweat:

Issue #161, opened 2016-08-14
Proposal: Source tree reorganization


jslee02:
I think the structure of FCL classes is well designed, but the structure of the source tree doesn't seem to accord with it. Here are some issues and suggestions I found while working on #150 and #154.

### Minimal classes in a header file
Some files have too many classes in them. This makes hard to learn FCL from navigating the source tree and to maintain the source code. Also, they are vulnerable to circular dependency issues as FCL will be templatized (#154). I propose splitting those files and having minimal classes (ideally single class) in a header file. One exception would be having specialized template classes in the same file of the template class.

### Reorganizing source tree structure
The top level directories and files are scattered. I suggest grouping them as `common`, `math`, `object`, `narrowphase`, and `broadphase`. Here (or see [this](https://github.com/flexible-collision-library/fcl/tree/reorganization/include/fcl)) is a source tree I imagine (bold denotes directory):

* **common**
  * `deprecated.h`, `profiler.h`, `time.h`, `types.h`, `warning.h`
* **math**
  * **bv**: Bounding volumes such as `AABB.h` and `OBB.h`
  * **motion**: Motions used for continuous collision detection
    * **taylor_model**
  * **sampler**
  * `constants.h`, `geometry.h` (utilities), `rng.h`, `triangle.h`, `variance3.h`
* **object**
  * **geometry**
    * **shape**: Shapes (e.g., `box.h`, `capsule.h`)
    * **bvh**: `BVH_model.h` and related files
    * **octree**
    * `collision_geometry.h`
  * `collision_object.h`, `continuous_collision_object.h`
* **narrowphase**
  * **detail**: Algorithm selectors and under-hood algorithms for collision geometries
    * **traversal**: Traversal nodes of BVH for different combination of collision geometries (a lot)
    * Algorithm selectors for primitive shapes: `gjk_solver_(indep/libccd).h` (could be renamed properly)
    * Algorithm table for collision geometries: `(collision/distance/conservative_advancement)_func_matrix.h`
  * Main interface for narrowphase queries: `collision.h`, `distance.h`, `continuous_collision.h`
  * Options and result data for the narrowphase queries: `(collision/distance/continuous_collision)_(request/result).h`
  * Basic data structures for the narrowphase queries: `contact.h`, `contact_point.h`, `cost_source.h`
* **broadphase**
  * **detail**: Internal data structure for the broadphase algorithms
  * broadphase related files

### Lower cases for directory names
[...]

PR #163, opened 2016-08-15
Reorganize source tree

jslee02:
This PR reorganize the FCL source tree as proposed #161.


jslee02:
I think the [build error (virtual memory exhausted: Cannot allocate memory)](https://travis-ci.org/flexible-collision-library/fcl/jobs/153285440#L770) is due to a huge amount of code generation for template objects. This should be fixed by #165.

PR #172, opened 2016-09-07
Exact negative distance computation using additional collision checking

jslee02:
For distance query, FCL seems to always (or in most case) return -1 distance when the two objects are in collision. Also, the comment of `DistanceResult::min_distant` is not clear whether this behavior was intentional or not.

This pull request adds a flag to `DistanceRequest` that indicates whether the caller wants the exact distance even when it's a negative value. If the flag is set to true and the distance is negative (objects are in collision), FCL will perform an additional collision checking **within** the distance routine to compute the exact negative distance that is the negated maximum penetration depth.

The default value of the flag is false, and this change adds only an overhead of one if-statement for the flag checking when the option is disabled.


sherm1:
Looks great, @jslee02. It might be good to mention how the penetration depth is calculated. I believe it is not always unambiguous so it would be good to know what method is being applied.

jslee02:
I'm not sure if describing the penetration depth calculation methods here (I guess you mean ) is a good idea because it varies in the geometry types. FCL has various penetration (also contact point) calculation methods such as for mesh-mesh (at the end it's triangle-triangle), mesh-primitive, mesh-octree, primitive-primitive, and so on. Furthermore, even for primitive-primitive, it could use the convexity-based methods or different method if there is a special algorithm only for the primitive shape pair (e.g., sphere-sphere, plane-sphere, and so on).

sherm1:
I see your point -- sounds good to me!

PR #181, opened 2016-09-26
Attempt to find libccd using find_package

jamiesnape:
For instance, if a developer has a `FindCcd.cmake` file or if danfis/libccd#24. is merged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,7 +79,9 @@ else()
 endif()
 
 # Find libccd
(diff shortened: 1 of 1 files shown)


jslee02:
This looks good to me.

Slightly off the topic: Shouldn't we match up the cases for the library name with the file name like `Findccd.cmake`?

jamiesnape:
Sure, though CMake itself is inconsistent. In practice, either will work.

Issue #138, opened 2016-06-29
Wrong value using Distance and Collision - FCL units?


dblanm:
Hello everyone!

I am testing the distance and collision of for cylinders. This cylinders modelate a robot of which I have the position and orientation. This is a part of the code:

```c++
typedef std::shared_ptr <fcl::CollisionGeometry> CollisionGeometryPtr_t;
CollisionGeometryPtr_t cylinder1 (new fcl::Cylinder(0.014,0.50));
CollisionGeometryPtr_t cylinder2 (new fcl::Cylinder(0.01,0.35));
//Set the transforms
fcl::Transform3f tf_robot_1, tf_robot_2;
// Set the orientation and position from Euler angles
Q.fromEuler(roll,pitch,yaw);
//... here is more code

// Define the collision Objects
fcl::CollisionObject link1 (cylinder1,tf_robot_1);
fcl::CollisionObject link2 (cylinder2,tf_robot_2);
//Define the requests and results
fcl::DistanceRequest distanceRequest(true);
fcl::DistanceResult distanceResult;
//For each of the links defined I do this
distanceResult.clear();
 FCL_REAL checkDistance1 = fcl::distance(&link1,&link2, distanceRequest, distanceResult);
```
However, I always check in gazebo if the collision exits and the distance, and both are usually wrong when the position (calculated using the joints degrees) is not in the exact point where the other cylinder is. 
I am defining the radio of the cylinders as 14 cm and 10 cm, or are  the units are different like inches? Otherwise what could be the problem? I checked if the position, and orientation calculated from Eulers, cuaternions and they were ok.

Thanks in advance!

Edit: Measurement distance was ok in FCL, clossing the issue.

PR #182, opened 2016-09-27
Attempt to find octomap using find_package

jamiesnape:
Looks slightly awkward due to inconsistent casing of some variables.


CMakeLists.txt
```diff
@@ -122,11 +123,13 @@ if(FCL_WITH_OCTOMAP)
           set(OCTOMAP_LIBRARIES "octomap;octomath")
       endif()
   endif()
-  if (OCTOMAP_FOUND OR (OCTOMAP_INCLUDE_DIRS AND OCTOMAP_LIBRARY_DIRS))
-    string(REPLACE "." ";" VERSION_LIST ${OCTOMAP_VERSION})
-    list(GET VERSION_LIST 0 OCTOMAP_MAJOR_VERSION)
-    list(GET VERSION_LIST 1 OCTOMAP_MINOR_VERSION)
-    list(GET VERSION_LIST 2 OCTOMAP_PATCH_VERSION)
+  if(OCTOMAP_INCLUDE_DIRS AND OCTOMAP_LIBRARY_DIRS AND OCTOMAP_VERSION)
+    if(NOT OCTOMAP_MAJOR_VERSION AND NOT OCTOMAP_MINOR_VERSION AND NOT OCTOMAP_PATCH_VERSION)
```
jslee02: I think [the upstream change #137](https://github.com/OctoMap/octomap/pull/137) should be merged first because we couldn't reach here without `OCTOMAP_VERSION`. Even it's merged this still wouldn't work if an octomap without the change is installed.

Possible option I can think of now is we enforce the minimum required octomap version to 1.8.1 (or the minimum version that includes the version variable) for CMake config mode like: `find_package(octomap 1.8.1 QUIET)` .

CMakeLists.txt
```diff
@@ -108,10 +108,11 @@ link_directories(${CCD_LIBRARY_DIRS})
 option(FCL_WITH_OCTOMAP "octomap library support" ON)
 set(FCL_HAVE_OCTOMAP 0)
 if(FCL_WITH_OCTOMAP)
-  if(PKG_CONFIG_FOUND)
+  find_package(octomap QUIET)
+  if(NOT OCTOMAP_INCLUDE_DIRS AND NOT OCTOMAP_LIBRARY_DIRS AND PKG_CONFIG_FOUND)
```
jslee02: Wouldn't it be simpler to use `octomap_FOUND` here? Maybe you used `OCTOMAP_INCLUDE_DIRS AND` and `OCTOMAP_LIBRARY_DIRS` because of the case inconsistency? If so, it would be good to leave a short note here why we don't use `XXX_FOUND` for maintenance in the future. I think the case inconsistency should be fixed in the upstream, though.

jslee02: Related upstream issue #138 (https://github.com/OctoMap/octomap/issues/138)

jamiesnape: True, though this would allow you to specify `OCTOMAP_INCLUDE_DIRS` and `OCTOMAP_LIBRARY_DIRS` as cache variables if you are using an old version without the necessary config.

CMakeLists.txt
jamiesnape: What would you do here? Remove the logic as redundant?

https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/config.h.in#L59-L73

CMakeLists.txt
```diff
@@ -122,11 +123,13 @@ if(FCL_WITH_OCTOMAP)
...
       endif()
   endif()
-  if (OCTOMAP_FOUND OR (OCTOMAP_INCLUDE_DIRS AND OCTOMAP_LIBRARY_DIRS))
-    string(REPLACE "." ";" VERSION_LIST ${OCTOMAP_VERSION})
-    list(GET VERSION_LIST 0 OCTOMAP_MAJOR_VERSION)
-    list(GET VERSION_LIST 1 OCTOMAP_MINOR_VERSION)
-    list(GET VERSION_LIST 2 OCTOMAP_PATCH_VERSION)
+  if(OCTOMAP_INCLUDE_DIRS AND OCTOMAP_LIBRARY_DIRS AND OCTOMAP_VERSION)
+    if(NOT OCTOMAP_MAJOR_VERSION AND NOT OCTOMAP_MINOR_VERSION AND NOT OCTOMAP_PATCH_VERSION)
+      string(REPLACE "." ";" VERSION_LIST "${OCTOMAP_VERSION}")
```
jamiesnape: Note that without `OCTOMAP_VERSION` there is currently an error here in due to the lack of quotes around `${OCTOMAP_VERSION}`

CMakeLists.txt
jslee02: The logic is still required. What I meant was that we enforce the minimum required octomap version _only_ when we attempt to find octomap using `find_package`.

CMakeLists.txt
jslee02: My concern was the case that octomap is installed and found by `find_package`, but FCL doesn't use it only because `OCTOMAP_VERSION` is not defined (due to `octomap-config.cmake` doesn't define it).

CMakeLists.txt
jslee02: The logic is still required. What I meant was that we enforce the minimum required octomap version _only_ when we attempt to find octomap using `find_package`.

jamiesnape: The current second (non-pkg-config) way  of finding octomap does not define OCTOMAP_VERSION. What do you currently do there?

CMakeLists.txt
jamiesnape: The configure will error out before that stage, so you need quoting and/or an `OCTOMAP_VERSION` guard.

jamiesnape: If version is not defined, I guess I could still call `pkg-config`, which at least fixes that scenario. Sound good?

jamiesnape: For platforms with `pkg-config` the previous behavior sound now be preserved with the current state of octomap.

jslee02: Sounds good!

Issue #186, opened 2016-11-08
Error: template argument required for 'class AABB'


ytzhao:
Hi, 
I'm using FCL for manipulator collision check, this library is very new to me. I followed the instructions to install this library on Ubuntu16.04 without ROS. I used 'general_test.cpp' for testing in Codeblocks with following build message:
'error: template argument required for 'class AABB''.

I've already included the library path and .h files in the project. I had several tries and found that if I commented <fcl/shape/geometric_shapes.h>, <fcl/shape/geometric_shapes_utility.h>, <fcl/narrowphase/narrowphase.h> these three files, there's no compiling error. 

Once I included either of these three header files, the compiling error appears. I traced down into these files and found that there are two AABB.h files in the library. One is <fcl/BV/AABB.h> and another is <fcl/math/bv/AABB.h>. The later file is written with 'template<typename S>' while the former one is not. It seems that if the former file is included, there's always the same error. So what's the different use between these two files? Is the one with template the newest version?

Anyone have ideas for solving this problem?

Thanks! :-)

jslee02:
`general_test.cpp` hasn't been updated for the new API (FCL 0.6). It was an oversight. I'll create a PR that updates the test.

The templated one is the newest version. I guess the non-templated one is the old version that's installed in the system directory.

ytzhao:
Thank you! Anyway, I reinstalled this library with more stable version 0.5, OpenRAVE and OMPL depend on the former version.

PR #188, opened 2016-11-10
Update general_test.cpp

jslee02:
This PR updates `general_test.cpp` for the new API, which should address #186.

I think this test could be replaced by [this](https://github.com/flexible-collision-library/fcl/blob/master/test/test_fcl_geometric_shapes.cpp#L709). If this makes sense then we could remove this test.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -48,6 +48,7 @@ set(tests
     test_fcl_collision.cpp
     test_fcl_distance.cpp
     test_fcl_frontlist.cpp
+    test_fcl_general.cpp
     test_fcl_geometric_shapes.cpp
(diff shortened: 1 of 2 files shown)


jslee02:
I think it's because of that this test checks for both of `float` and `double` types while the other tests don't. It generates template code for `float` type, but the generated code is not tested by other tests. I just disabled other tests for `float` not to double the testing time.

I'm not sure if we want to test for `float` type besides `double`. If it's not, we could just disable the `float` test of `fcl_test_general.cpp`.

jslee02:
I believe AppyVeyor test will pass if we re-run it. I'll fix it if it still exists after this PR is merged.

Issue #184, opened 2016-10-30
lots of compilation errors


xendian:
I am using visual studio 20113 and getting lots of errors on ;
Error   10  error C2146: syntax error : missing ';' before identifier 'S'  in  ....\fcl-master\src\broadphase\broadphase_collision_manager.cpp

due to in .....\fcl-master\include\fcl\math\constants.h line number 49

I am gettin 591 errors due to this.. Any body has the same issue?

sherm1:
I believe you need VS 2015 to compile successfully.

xendian:
Thanks!!.I will try..it seems visual 2013 compiler has a problem with  the constexpr

xendian:
yes, visual studio 2015 solved the issues.

PR #189, opened 2016-11-10
Explicitly require VS2015 or greater

jslee02:
Related issue: #184

### DIFF
--- a/CMakeModules/CompilerSettings.cmake
+++ b/CMakeModules/CompilerSettings.cmake
@@ -3,15 +3,15 @@ if(CMAKE_COMPILER_IS_GNUCXX)
     add_definitions(-std=c++11 -W -Wall -g -Wextra -Wpedantic -Wno-missing-field-initializers -Wno-unused-parameter)
     if(FCL_TREAT_WARNINGS_AS_ERRORS)
         add_definitions(-Werror)
-    endif(FCL_TREAT_WARNINGS_AS_ERRORS)
-endif(CMAKE_COMPILER_IS_GNUCXX)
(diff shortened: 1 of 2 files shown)


CMakeModules/CompilerSettings.cmake
```diff
@@ -27,12 +27,15 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
...
 # Visual Studio
-if(MSVC OR MSVC90 OR MSVC10)
+if(MSVC)
+    if(MSVC_VERSION VERSION_LESS 1900)
+        message(FATAL_ERROR "${PROJECT_NAME} requires Visual Studio 2015 or greater.")
+    endif()
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHsc /MP /W1 /bigobj")
     if(FCL_TREAT_WARNINGS_AS_ERRORS)
         add_definitions(/WX)
     endif(FCL_TREAT_WARNINGS_AS_ERRORS)
```
sherm1: Consider `endif()` for consistency.

jslee02: Actually, I don't know which one is the convention here. Some `endif()` contains parameter while some don't.

sherm1: I agree! `endif()` is always better, IMO. That is very clear when the original condition is some complicated expression!

jslee02:
I removed conditions from `endif()`. It seems `endif()` is the convention according to the `CMakeLists.txt` files of FCL.

PR #196, opened 2017-01-12
Clean up finding external dependencies and use imported targets where available

jamiesnape:
* Uses the output of `pkg-config` as input to `find_library()` and `find_path()` calls so that CMake is always provided with the absolute paths to libraries, as is recommended. Also avoids the need to use the ambiguous and frowned upon `link_directories()`.
* Uses `IMPORTED` targets for external libraries where available. Recent versions of Eigen, libccd, and OctoMap all provide these. If `IMPORTED` targets are not available `*_INCLUDE_DIRS` and `*_LIBRARIES` variables are used, as previously.
* Populates `INTERFACE_INCLUDE_DIRECTORIES` for `fcl` with the locations of all necessary external headers via either `target_link_libraries()` for `IMPORTED` targets or `target_include_directories()` otherwise. The locations of the headers for libccd and OctoMap were missing before. Also removes redundant calls to `include_directories()`.
* Adds `eigen3`, `libccd`, and `octomap`, if necessary, to the `Requires` field of `fcl.pc`. All were missing before. Also adds the missing `Description`.
* Removes the `-g` from the compiler flags always passed to GCC. The `Debug` and `RelWithDebInfo` CMake build types exist for this purpose.


jamiesnape:
Note that master is currently broken: https://travis-ci.org/flexible-collision-library/fcl/builds/188952125.

CMakeLists.txt
```diff
@@ -99,83 +99,134 @@ if(FCL_COVERALLS)
...
+#
+# If Eigen3 is not found, manually set the cache variable EIGEN3_INCLUDE_DIR
+#===============================================================================
+find_package(Eigen3 3.0.5 QUIET CONFIG)
+
+# If Eigen3Config.cmake is not found, use the FindEigen3.cmake module
+if(NOT Eigen3_FOUND)
+  find_package(Eigen3 3.0.5 QUIET MODULE)
+  set(Eigen3_FOUND ON)
+endif()
```
jslee02: Why do we need to find packages in config mode and module mode explicitly? I though `find_package()` is for it. Is this for searching the config file first? If so, why is that necessary?

CMakeLists.txt
```diff
@@ -99,83 +99,134 @@ if(FCL_COVERALLS)
...
+  endif()
+
+  if(CCD_INCLUDE_DIR AND CCD_LIBRARY)
+    set(CCD_INCLUDE_DIRS "${CCD_INCLUDE_DIR}")
+    set(CCD_LIBRARIES "${CCD_LIBRARY}" "${M_LIBRARY}")
+    set(ccd_FOUND ON)
+
+    mark_as_advanced(CCD_INCLUDE_DIR CCD_LIBRARY)
   endif()
 endif()
```
jslee02: I would like to suggest encapsulating this logic (and for other dependencies as well) as a separate custom module like `FindCCD.cmake`. That would make this main CMakeLists.txt cleaner.

CMakeLists.txt
jamiesnape: Because CMake uses `MODULE` mode first and falls back to `CONFIG` mode, which unfortunately is the opposite that we would want.

jslee02: Why do we want to search in that order then?

jamiesnape: The config file is maintained by Eigen will adapt to any changes that they make. Also it defines imported targets. The reason that `find_package` defaults to the other direction so as to not unknowingly break existing projects.

CMakeLists.txt
jslee02: We could do this later if needed. I'm merging this PR now.

PR #204, opened 2017-01-27
Fix more compiler warnings

jamiesnape:
PTAL @jslee02.


CMakeModules/CompilerSettings.cmake
```diff
@@ -1,14 +1,14 @@
 # GCC
 if(CMAKE_COMPILER_IS_GNUCXX)
-    add_definitions(-std=c++11 -W -Wall -g -Wextra -Wpedantic -Wno-missing-field-initializers -Wno-unused-parameter)
+    add_definitions(-std=c++11 -W -Wall -g -Wextra -Wpedantic -Wno-unused-parameter)
```
jslee02: I believe we could remove `-Wno-unused-parameter` as well. One case could be a problem I can think of is using variables that are only used in assertions. For that case, we could use a macro to suppress the warning as [we did in DART](https://github.com/dartsim/dart/pull/646).

jamiesnape: Almost, I have another PR for that. Touches something like 36 files.

jamiesnape: See #205.

PR #208, opened 2017-02-07
added missing typename.

Remotion:
replaced not with ! because intel compiler do not want to compile this.
replaced integrate(double dt) with integrate(S dt) to allow fcl compile with float.


include/fcl/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+cmake_minimum_required(VERSION 3.6)
```
jslee02: It would be good to remove this line unless there is a compelling reason we need this. FCL is supposed to support Ubuntu 14.04 LTS, and the [CMake version on Ubuntu 14.04 LTS is 2.8.12](http://packages.ubuntu.com/trusty/cmake).

Remotion:
CMakeLists.txt  was changed by accident. 

But I already pushed other changes before "Moving easy to inline code from *.cpp to header files."

What do you think?

jslee02:
Why do we need this explicit inlining? As I know, the inlining decision is implicitly made by compilers if it's needed for optimization even it's not in the same compilation unit. I believe defining definition in the header file would only make headers less readable.

Remotion:
It is not about inlining it is because of One Definition Rule. 

It is not needed for methods declared inside of class/struct like BV_node_base.
But if it is done outside and is not a template then inline is required. 


> I believe having definition in the header file would only makes headers less readable.
Well this is only a couple of lines, not that much and without it FLC is for sure hot header only because it requires *.cpp files to compile and link.

Of corse another way would be to put such code into plane-inl.h files.

jslee02:
I think there is a miscommunication here. FCL is not a header-only project yet. Sorry if I brought this confusion by the previous comment.

We've templatized FCL for the scalar type so that we can use AutoDiffScalar to get derivatives (see @sherm1's [suggestion](https://github.com/flexible-collision-library/fcl/issues/96#issuecomment-234699398) and related PRs: #154, #165). The templatizing was necessarily applied to most of the code, so it seems easy to make FCL an (optional) header-only project. However, we haven't made a decision on that yet. It would be good to open a new issue about this to discuss with other contributors and postpone the inlining of this PR until then.

Issue #106, opened 2016-03-26
Incorrect contact point computation of mesh-mesh collision


jslee02:
[Here](https://github.com/flexible-collision-library/fcl/commit/13a1f78e32ccdf26d1da3a3205480e860b456413) is a simple test to show the incorrect contact point computation in mesh-mesh collision I mentioned in #97. 

In the test, there are two meshes generated from [`Box`](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/shape/geometric_shapes.h#L76) as shown in below figure (rendered by DART). The two meshes are in face contact. We would expect four contact points should be at the four corners of the face of green box touching with the red box (or at least not out of the face). However, the results showed that the contact points (small blue spheres in the figure) are more than four (30 points and some of them are duplicates) and even placed at the corner of the red box.

![image](https://cloud.githubusercontent.com/assets/4038467/14061657/fc4c52e2-f35b-11e5-9d73-961110cb706d.png)

It seems that contact points are always placed at the vertices of mesh triangles rather than the exact contact point of the two colliding triangles.

This bug is crucial and easily results in unrealistic behavior in dynamics simulation. For example, the yellow box in the below figure is hanging in the air because of the ghost contact points at the right bottom corners.

![image](https://cloud.githubusercontent.com/assets/4038467/14061778/abd5f724-f35f-11e5-9430-8d685cf2bf7a.png)

sherm1:
Very convincing! Is this specific to box primitives or does it apply in general to contact points for any meshes?

jslee02:
I believe (without proof) this is not specific to box primitives (to clarify, mesh generated from box was tested rather than the original box primitive), but this is more noticeable when the triangle is big. For box mesh case, each triangle is the half size of the square face.

Here is another quick test of box-cylinder meshes.
![image](https://cloud.githubusercontent.com/assets/4038467/14061936/41829e22-f364-11e5-914d-e35fdbf9f227.png)

There is a weird contact point at the middle of the yellow cylinder. It seems the three contact points are the vertices of one triangle of the yellow cylinder colliding with the red box, and the weird contact point is reported only because it's on the triangle.

anna-seppala:
Has somebody worked on fixing this issue? I have similar problems when generating contact points for two OT_BVH objects with BV_OBBRSS node type; many of the contact points are duplicates and some are wildly scattered around the two objects in collision. That being said, I have not pulled the newest changes on but am working on a forked project from some time ago.

[...]

PR #211, opened 2017-04-08
Handle collisions between coplanar triangles

a-price:
Addresses #106, at least as far as exercised by the test suite. @jslee02, can you verify for the example cases you provided, or add tests for any cases that still fail?


test/CMakeLists.txt
```diff
@@ -30,14 +30,15 @@ add_definitions(-DGTEST_DONT_DEFINE_ASSERT_GT=1)
 
 execute_process(COMMAND cmake -E remove_directory ${CMAKE_BINARY_DIR}/test_results)
 execute_process(COMMAND cmake -E make_directory ${CMAKE_BINARY_DIR}/test_results)
-include_directories(${GTEST_INCLUDE_DIRS})
+include_directories(SYSTEM ${GTEST_INCLUDE_DIRS})
```
avalenzu: Why is this change necessary?

test/test_fcl_coplanar_triangles.cpp
```diff
@@ -0,0 +1,243 @@
...
+  {
+    for (int j = 0; j < numContacts; ++j)
+    {
+      C(i,j) = (expected[i]-actual[j]).squaredNorm();
+    }
+  }
+
+  fcl::test::Hungarian<Scalar> H(C);
+
+  std::vector<int> permutation = H.getAssignment();
```
avalenzu: Could you achieve the same effect here by sorting `expected` and `actual` with `std::sort` and a custom comparison? Adding the Hungarian algorithm just to get the corresponding points seems like overkill. The following comparison should work, though `expected` and `actual` would need to be modified to be `std::vector<Vector3<Scalar>>`. Of cource, it's also quite possible that I'm missing something.
```c++
template <typename Scalar>
bool comparePoints(const Vector3<Scalar>& v1, const Vector3<Scalar>& v2)
{
  for (int i = 0; i < 3; ++i) {
    if (v1[i] < v2[i]) {
      return true;
    } else if (v1[i] > v2[i]) {
      return false;
    }
  }
  // In this case v1 == v2
  return false;
}
```

avalenzu:
@a-price, just want to check if you've seen my code review.

test/CMakeLists.txt
a-price: Including directories as SYSTEM directories suppresses compiler warnings in includes from dependencies. More info [here](https://cmake.org/cmake/help/v3.0/command/include_directories.html) and [here](https://stackoverflow.com/a/26297802).

a-price:
@avalenzu Hey, thanks for the feedback. I'm just now getting back to this. I've pushed a couple of the requested changes, and added comments on the things I haven't switched. Sorry about the delay.

test/test_fcl_coplanar_triangles.cpp
a-price: Yeah, this is probably overkill. I had that code lying around from something else, so I figured it was easiest to just do the matching the "right" way, but only have the class in the test methods so as to not expose it to the main body of code. Is it a problem to have it there?

PR #213, opened 2017-05-01
Hierarchy tree bottom-up building: initialize iterators

JenniferBuehler:
A segmentation fault is caused if the iterators in the hierarchy tree bottomup() method aren't initialized. It can happen that the minimum spanning BV cannot be found, for example if one of the BV's has a nan value size (or same for inf value).

While this would be an error arising earlier on, e.g. when updating the AABBs, it would be good to at least have the iterators here initialized to do anything at all.

We need this fix for FCL 0.5 or lower (compatible with DART 6.1). Of course the same fix could / should be applied to the master branch. Please let me know what you think about this, and then I can also post a PR to master and/or FCL 0.3 and 0.4.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/213)

### DIFF
--- a/include/fcl/broadphase/hierarchy_tree.hxx
+++ b/include/fcl/broadphase/hierarchy_tree.hxx
@@ -280,11 +280,21 @@ void HierarchyTree<BV>::bottomup(const NodeVecIterator lbeg, const NodeVecIterat
   while(lbeg < lcur_end - 1)
   {
     NodeVecIterator min_it1, min_it2;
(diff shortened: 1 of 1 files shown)


bmagyar:
I can confirm that this fix solves some segfault crashes within MoveIt too. (ROS Kinetic)
I am using a more advanced planner which seems to have challenged `fcl` when self-collision checking

The issue for me first appeared 2017 August, now resurfaced and with my current setup appeared 50%, glad I found this PR!

PR #215, opened 2017-05-31
Return the correct nearest points from `distance(...)`

avalenzu:
This PR fixes issues in both the CCD-based and the independent GJK solvers, which caused `distance` to return erroneous values for the nearest points.

- CCD-base solver: Currently the most recently computed support points are returned as the nearest points, which is incorrect. The nearest points lie in the convex-hull of the points contained in the `simplex` variable. The points on the original shapes can be computed from the `simplex` and the nearest point to the origin on the Minkowski difference, `dir`.
- Independent solver: Currently, `p2` is not properly transformed to its shape's (`shape1`) frame. `toshape0` converts points from `shape1` frame to `shape0` frame. The points extracted from the GJK simplex are in `shape0` frame, so the inverse of `toshape0` needs to be applied.

The `test_fcl_capsule_box_*` tests contained lines that failed due to the issues described above. This PR uncomments those lines and modifies the tests to exercise both solvers.

This PR also adds a `distance_tolerance` member to `fcl::DistanceRequest`, which corresponds to the `distance_tolerance` and `gjk_tolerance` members in `GJKSolver_libccd` and `GJKSolver_indep` respectively. @jslee02, is this an appropriate way to expose this knob to the user? I didn't use `abs_err` because the default for that was 0.0, whereas the defaults for `distance_tolerance` and `gjk_tolerance` are 1e-6.

### DIFF
--- a/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
+++ b/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
@@ -1025,19 +1025,42 @@ static void extractClosestPoints(ccd_simplex_t* simplex,
       //
       // p = A + s*AB, 0 <= s <= 1
       // p - A = s*AB
+      ccd_vec3_t AB;
+      ccdVec3Sub2(&AB, &(simplex->ps[1].v), &(simplex->ps[0].v));
+
+      // This defines three equations, but we only need one. Taking the i-th
+      // component gives
       //
-      // This defines three equations, but we only need one. Take the x
-      // component
-      //
-      // p_x - A_x = s*AB_x
+      // p_i - A_i = s*AB_i.
       //
       // Thus, s is given by
       //
-      // s = (p_x - A_x)/(B_x - vA_x)
-      double A_x{ccdVec3X(&(simplex->ps[0].v))};
-      double B_x{ccdVec3X(&(simplex->ps[1].v))};
-      double p_x{ccdVec3X(p)};
-      double    s{(p_x - A_x) / (B_x - A_x)};
+      // s = (p_i - A_i)/AB_i.
+      //
+      // To avoid dividing by an AB_i ≪ 1, we choose i such that |AB_i| is
+      // maximized
+      ccd_real_t abs_AB_x{std::abs(ccdVec3X(&AB))};
+      ccd_real_t abs_AB_y{std::abs(ccdVec3Y(&AB))};
+      ccd_real_t abs_AB_z{std::abs(ccdVec3Z(&AB))};
+      ccd_real_t s{0};
+
+      if (abs_AB_x >= abs_AB_y && abs_AB_x >= abs_AB_z) {
+        ccd_real_t A_x{ccdVec3X(&(simplex->ps[0].v))};
+        ccd_real_t AB_x{ccdVec3X(&AB)};
+        ccd_real_t p_x{ccdVec3X(p)};
+        s = (p_x - A_x) / AB_x;
+      } else if (abs_AB_y >= abs_AB_z) {
+        ccd_real_t A_y{ccdVec3Y(&(simplex->ps[0].v))};
+        ccd_real_t AB_y{ccdVec3Y(&AB)};
+        ccd_real_t p_y{ccdVec3Y(p)};
+        s = (p_y - A_y) / AB_y;
+      } else {
+        ccd_real_t A_z{ccdVec3Z(&(simplex->ps[0].v))};
+        ccd_real_t AB_z{ccdVec3Z(&AB)};
+        ccd_real_t p_z{ccdVec3Z(p)};
+        s = (p_z - A_z) / AB_z;
+      }
+
       if (p1)
       {
         // p1 = A1 + s*A1B1
@@ -1060,15 +1083,15 @@ static void extractClosestPoints(ccd_simplex_t* simplex,
   else // simplex_size == 3
   {
     // Closest points lie in the triangle defined by the points in the simplex
-    ccd_vec3_t AB, AC, n, v_prime, AB_cross_v_prime, AC_cross_v_prime, AB_cross_AC;
+    ccd_vec3_t AB, AC, n, v_prime, AB_cross_v_prime, AC_cross_v_prime;
     // Let the triangle be defined by points A, B, and C. The triangle lies in
-    // the plane that passes through A, whose normal is given by
+    // the plane that passes through A, whose normal is given by n̂, where
     //
-    // n = (AB x AC) / || AB x AC ||
+    // n = AB × AC
+    // n̂ = n / ‖n‖
     ccdVec3Sub2(&AB, &(simplex->ps[1].v), &(simplex->ps[0].v));
     ccdVec3Sub2(&AC, &(simplex->ps[2].v), &(simplex->ps[0].v));
     ccdVec3Cross(&n, &AB, &AC);
-    ccdVec3Normalize(&n);
 
     // Since p lies in ABC, it can be expressed as
     //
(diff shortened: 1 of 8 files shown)


avalenzu:
@jslee02, do you have any insights as to why libccd would return different results on Linux than it does on Windows and OSX? `test_fcl_capsule_box_1` passes on Linux, but fails on the other two platforms.

jslee02:
> @jslee02, do you have any insights as to why libccd would return different results on Linux than it does on Windows and OSX? test_fcl_capsule_box_1 passes on Linux, but fails on the other two platforms.

One possible reason would be that `fcl` uses different versions of `libccd` for the tests depending on platforms? [`fcl` uses 1.5 on Linux](https://launchpad.net/~libccd-debs/+archive/ubuntu/ppa/+packages) and 2.0 for the other two platforms. You could test this by manually installing `libccd` 2.0 on Linux.

avalenzu:
> You could test this by manually installing libccd 2.0 on Linux for sure.

Hmmm. I've been using the `libccd-dev` package from the Xenial apt repositories (which uses the libccd 2.0 sources) on my machine, and all the tests pass for me.

jslee02:
I realized that `libccd-dev` in the Ubuntu apt repository was built with double precision while macOS uses single precision. If you see [the debian file](https://packages.ubuntu.com/xenial/libccd-dev), then it passes `-DCCD_DOUBLE:BOOL=True` option but [the homebrew formula](https://github.com/Homebrew/homebrew-science/blob/master/libccd.rb) just uses the default setting that is single precision.

If this is the cause of the issue, then we might want to add a precision option to the homebrew formula.

sherm1:
Is FCL really useful in single precision? Likely it could at least default to double everywhere, but it might be worth simply declaring that only double is supported, especially if there are unit tests that won't pass in float.

edrumwri:
I think Jeongseok was talking about the default build option for libccd,
not FCL.


--------------------------------
Evan Drumwright
Senior Research Scientist
http://positronicslab.github.io
Toyota Research Institute
Palo Alto, CA

avalenzu:
@jslee02, it looks like that's the issue. I verified that `test_fcl_capsule_box_1` passes on OSX when libccd is built with double precision, and fails when libccd is build with single precision.

avalenzu:
Thanks for finding the precision issue, @jslee02!

I think we should update the homebrew formula to use double precision. @jslee02 do you have experience editing homebrew formulae? If so, could you put in a pull request with that change? I haven't the slightest idea how to go about testing changes to homebrew configuration. I'm willing to learn, but if you already have that know-how, it would be awesome if you could put that in :smiley: 

Additionally, is there a way to require at build time that libccd was built with double precision (i.e. that `ccd_real_t` is `double`)? As @sherm1 mentioned, it doesn't really make sense to allow building in a configuration that's known to yield wrong answers.

include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
```diff
@@ -1005,6 +1005,130 @@ static int __ccdEPA(const void *obj1, const void *obj2,
...
+  else if (simplex_size ==2)
+  {
+      // Closest points lie on the segment defined by the points in the simplex
+      // Let the segment be defined by points A and B. We can write p as
+      //
+      // p = A + s*AB, 0 <= s <= 1
+      // p - A = s*AB
+      //
+      // This defines three equations, but we only need one. Take the x
+      // component
```
SeanCurtis-TRI: The quality of this solution degrades as the line segment becomes more perpendicular to the x-axis (i.e., where `B_x - A_x << 1`. Ideally, you'd want to pick the axis where `|B_i - A_i|` is biggest.

include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
```diff
@@ -1005,6 +1005,130 @@ static int __ccdEPA(const void *obj1, const void *obj2,
...
+      //
+      // p_x - A_x = s*AB_x
+      //
+      // Thus, s is given by
+      //
+      // s = (p_x - A_x)/(B_x - vA_x)
```
SeanCurtis-TRI: typo?  `vA_x`

include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
```diff
@@ -1005,6 +1005,130 @@ static int __ccdEPA(const void *obj1, const void *obj2,
...
+    // n . (AC x p') = -s * n . (AB x AC)
+    double n_dot_AB_cross_AC{ccdVec3Dot(&n, &AB_cross_AC)};
+
+    // Therefore, s and t are given by
+    //
+    // s = -n . (AC x p') / n . (AB x AC)
+    // t =  n . (AB x p') / n . (AB x AC)
+    double s{-ccdVec3Dot(&n, &AC_cross_v_prime) / n_dot_AB_cross_AC};
+    double t{ccdVec3Dot(&n, &AB_cross_v_prime) / n_dot_AB_cross_AC};
+
```
SeanCurtis-TRI: I believe a little algebraic manipulation simplifies the math (making it cheaper and more robust -- no square roots and fewer floating point ops).

Starting with the following definitions (and focusing only on `t` with the understanding that it applies directly to `s`:
```
n = AB x AC
n̂ = n / ||n||
t = n̂⋅(AC x p') / (n̂⋅(AB x AC))
```
We get the following:
```
t = n̂⋅(AC x p') / (n̂⋅(AB x AC))                              - the given equation
t = (n / ||n||)⋅(AC x p') / ((n / ||n||)⋅(AB x AC))          - substitute for n̂
t = n⋅(AC x p') / (n⋅(AB x AC))                              - cancel 1 / ||n||
t = (AB x AC)⋅(AC x p') / ((AB x AC)⋅(AB x AC))              - substitute for n
t = (AB x AC)⋅(AC x p') / ||AB x AC||²                       - simplify
```

include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
```diff
@@ -1005,6 +1005,130 @@ static int __ccdEPA(const void *obj1, const void *obj2,
...
+      //
+      // p_x - A_x = s*AB_x
+      //
+      // Thus, s is given by
+      //
+      // s = (p_x - A_x)/(B_x - vA_x)
+      double A_x{ccdVec3X(&(simplex->ps[0].v))};
+      double B_x{ccdVec3X(&(simplex->ps[1].v))};
+      double p_x{ccdVec3X(p)};
+      double    s{(p_x - A_x) / (B_x - A_x)};
```
SeanCurtis-TRI: nit: extra space proceeding `s`

jslee02:
> @jslee02 do you have experience editing homebrew formulae? If so, could you put in a pull request with that change?

I think I could do that. Let me put that it to my list.

include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
avalenzu: Done.

include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
avalenzu: Fixed.

include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
avalenzu: Nice! Done. PTAL.

jslee02:
Submitted a PR: Homebrew/homebrew-science#5762

avalenzu:
I see that the homebrew PR went through! Is there a way to re-trigger the CI checks for this PR?

avalenzu:
@jslee02, am I correct in thinking that this PR is blocked by #216?

Also, could I get your take on the last paragraph of the PR description? Thank you!

jslee02:
Sorry for delayed response.

#216 is now merged. Could you merge master into this PR so that the CI checks are re-triggered as well?

jslee02:
> @jslee02, is this an appropriate way to expose this knob to the user?

It looks good to me. Adding a new parameter to `DistanceRequest` would break the API compatibility, but it would be okay because FCL 0.6.0 breaks it anyway by other changes.

avalenzu:
Is it possible that the cached libccd is still single precision on the x64 Windows machine?

jslee02:
I believe so. Let's refresh the cache by [removing](https://www.appveyor.com/docs/build-cache/#remove-cache-entry-from-build-config) it and adding it back (enabling the cache).

avalenzu:
OK, the appveyor/pr check passed once the cache was cleared. I've re-enabled the cache now.

jslee02:
Hm, it seems the cache isn't reset by doing so. Instead, I found another method ([ref1](https://www.appveyor.com/docs/build-cache/#cache-dependencies), [ref2](http://help.appveyor.com/discussions/questions/1310-delete-cache)). Could you change this line
```yaml
cache:
  - C:\Program Files\libccd
```
to
```yaml
cache:
  - C:\Program Files\libccd -> appveyor.yml
```
?

Thanks!

avalenzu:
@jslee02: appveyor/pr is finally happy! What's being checked in appveyor/branch?

jslee02:
I believe appveyor/branch checks for regular commits to the branch, but [it seems not to work sometimes](http://help.appveyor.com/discussions/problems/3644-pull-request-status-on-github-shows-three-appveyor-status-first-one-stays-on-yellow-forever). I just removed it from the requirements as suggested in the post.

Merging this! Thank you for the contribution!

PR #216, opened 2017-06-01
Build libccd with double precision on Windows

jslee02:
FCL fails to pass `test_fcl_capsule_box_1` test with `libccd` built with single precision, which was discovered in #215. To partially address the issue, this PR changes the CI build scripts for Windows to always use `libccd` built with double precision.

We might want to do similar work for macOS as well.

Unrelated change: `libccd` is now built in `build` directory.


jslee02:
@sherm1 Yeah, I updated the description accordingly. I created this PR without description in motion to quickly trigger the CI tests. :smile: 

[It seems there is a confliction with `build` directory only on Win32](https://ci.appveyor.com/project/mamoll/fcl/build/502/job/hqj1xj2yxm6leqe3#L864). Will resolve this soon.

PR #228, opened 2017-09-14
Approximate signed distance when distance is equal to 0

### LINKED ISSUE #221: Signed distance doesn't work on BVH Models.
According to the comment at https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/narrowphase/distance-inl.h#L140, there should be a work around in place to calculate negative distance for BVH meshes (even if the closest points aren't correct), but in practice, even `enabling request.enable_signed_distance = true` doesn't calculate signed distance, and the returned min_distance is always 0. 

I've put together a test demonstrating this [here](https://github.com/BryceStevenWilley/fcl/blob/master/test/test_fcl_signed_distance.cpp).

marc1706:
The min distance calculation on shapes that are not of type OT_GEOM is not supported and will return 0. However, the previous check was evaluating as false as 0 is clearly not less than static_cast<S>(0) and the workaround to have an approximate signed min distance was never executed.

Fixes #221 

Note: We incidentally worked around this bug ourselves with code that looked the same as the workaround in distance-inl.h.


wxmerkt:
What is currently blocking this PR? It needs to be rebased/updated and OS X Travis couldn't find libccd?

wxmerkt:
This fix actually introduces the possibility for a segfault when ``res`` is -1 which can happen if there is no intersection detected.

marc1706:
Although I find it highly disturbing that res can be a negative value without any intersections, I have now adjusted the pull request to make sure that getContact() does not get called if there are no contacts (I guess this is the potential segfault you were referring to).

saunair:
Any plan on merging this if this is a potential fix?

SeanCurtis-TRI:
Couple of quick thoughts -- and I'll look at this tomorrow in a bit more detail.

1) The apparent CI failure looks like historical garbage. It seems that the cause has already been rectified, but CI hasn't run on this PR since before the fix. If you push to this PR again, CI will re-trigger and I'm sure that failure will evaporate.
2) You need to rebase this on master to make it mergable (and this will likewise re-trigger CI).
3) We really need a unit test that fails without this patch but works *with* it.

Thanks

saunair:
@SeanCurtis-TRI . I should have posted that with a bit more detail. 
1. I used libccd 
2. In the distance test, I deliberately set the flags of signed_distance and nearest_pts to be true and also set a couple of transforms to be in penetration, but the distance returned is zero and the test still passes. 
    My point here was just that if we fix this issue, we can write a test cases that deliberately check negative values that are expected.  
3. Positive distances seemed right to me all the time in all cases with a without this change in the PR. 
4. I have a couple of tests that we can use if we fix this issue. 
The following snippet works with the fix
```
template<typename BV>
void sphere_mesh_signed_dist()
{
    using S = typename BV::S;
    Transform3<S> tf_1 = Transform3<S>{Translation3<S>(Vector3<S>(0.0, 0., 0.0))};
    Transform3<S> tf_2 = Transform3<S>{Translation3<S>(Vector3<S>(0.0, 0., 1.0))};

    Sphere<S> s1(1);
    Sphere<S> s2(1);

    BVHModel<BV> s1_rss;
    BVHModel<BV> s2_rss;
    s1_rss.bv_splitter.reset(new detail::BVSplitter<BV>(detail::SPLIT_METHOD_MEAN));
    s2_rss.bv_splitter.reset(new detail::BVSplitter<BV>(detail::SPLIT_METHOD_MEAN));

    generateBVHModel(s1_rss, s1, Transform3<S>::Identity(), 64);
    generateBVHModel(s2_rss, s2, Transform3<S>::Identity(), 64);
    DistanceRequest<S> request;
    request.enable_signed_distance = true;
    request.enable_nearest_points = true;
    request.gjk_solver_type = GST_LIBCCD;
    DistanceResult<S> result;
  
    distance(&s1_rss, tf_1, &s2_rss, tf_2, request, result);
    std::cout<<"sphere dist = "<<result.min_distance<<std::endl;
}
```

The following snippet moves the mesh along an axis and prints the pen-depths at every stage.
Here, we get values that range around 5e-5 to 0.001 and then back again till we come out of penetration. 
We should be getting a continuous noisy curve in a direction, but just seems random negative values instead. 
```
template<typename BV>
void mesh_to_mesh_signed_dist()
{
    using S = typename BV::S;
    double start = 0.1; 
    for(int i=0; i<200; i++)
{
    Transform3<S> tf_1 = Transform3<S>{Translation3<S>(Vector3<S>(0.0, 0., 0.0))};
    Transform3<S> tf_2 = Transform3<S>{Translation3<S>(Vector3<S>(0.0, start, 0.0))};
    start -= 0.005;
    std::vector<Vector3<S>> p1;
    std::vector<Triangle> t1;

    test::loadOBJFile(TEST_RESOURCES_DIR"/secret.obj", p1, t1);

    std::vector<Vector3<S>> p2;
    std::vector<Triangle> t2;

    test::loadOBJFile(TEST_RESOURCES_DIR"/secret.obj", p2, t2);
    BVHModel<BV> m1 = BVHModel<BV>();
    BVHModel<BV> m2 = BVHModel<BV>();
    m1.bv_splitter.reset(new detail::BVSplitter<BV>(detail::SPLIT_METHOD_MEAN));
    m2.bv_splitter.reset(new detail::BVSplitter<BV>(detail::SPLIT_METHOD_MEAN));

    m2.beginModel();
    m2.addSubModel(p2, t2);
    m2.endModel();

    m1.beginModel();
    m1.addSubModel(p1, t1);
    m1.endModel();
    DistanceRequest<S> request;
    request.enable_signed_distance = true;
    request.enable_nearest_points = true;
    request.gjk_solver_type = GST_LIBCCD;
    DistanceResult<S> result;
    result.clear();
    //detail::MeshDistanceTraversalNode<BV> node;
    //node.enable_statistics = true;
    distance(&m1, tf_1, &m2, tf_2, request, result);

    std::cout<<"dist = "<<result.min_distance<<std::endl;
}
}

```

saunair:
secret.obj is an internal object that I was using. 
I ran this over fcl:: rob.obj and got some very interesting results. Mostly looks like it works . 
rob

PR #233, opened 2017-10-11
 C++ visibility support (take II)

j-rivero:
Continues discussion and work done in #231. Reproducing the summary: 

This PR implements the [control of binary symbols visibility](https://gcc.gnu.org/wiki/Visibility) in FCL.

It adds a new `FCL_VISIBLE` preprocessor define to any class, struct or function that needs to be exposed publicly to third party software. Since the number of files in this library is not small, I've patched all what I found needed to compile the whole test suite. It would be good to use another third party software using FCL extensively to test that I did not forget about any of them.

The PR also adds the `FCL_HIDE_ALL_SYMBOLS` option (OFF by the default). This option covers a use case we found in Drake where a software uses FCL (linked statically) but does not expose it in the public API (headers). In this scenario the best option to avoid any possible conflict with other versions of fcl is to hide all the symbols from the public ABI.

Numbers before and after:

| Version | Number of symbols | libfcl size |
| ----------|--------------------------|-------------|
| current | 18942                      | 11922 Kb     |
|with visibility from this PR | 16530 | 11257Kb |
|no symbols | 157 |  9012Kb |

Summary: **13% symbols less** and **6% size less** . 

The PR will alter the ABI so a new bump in the FCL version prior to release it would be ideal.


j-rivero:
I think that all the compilers/options on Linux and Windows are now happy. Mac is failing due to a problem finding ccd (I think it is unrelated to this PR).

j-rivero:
@jslee02 I've been trying to compile this branch with Dart (master and tag 6.3.1) but there are compilation failures due to a change in headers location. Do you know if there is any branch ready in Dart ready to work with fcl current master?

jslee02:
@j-rivero https://github.com/dartsim/dart/pull/873 should work with this PR. I tested it against FCL 0.5 as well.

CMakeLists.txt
```diff
@@ -229,6 +233,11 @@ else()
   message(STATUS "FCL does not use OctoMap (as requested)")
 endif()
 
+# Need to include directory to find export file
+include_directories(${PROJECT_BINARY_DIR}) 
+
+install(FILES ${PROJECT_BINARY_DIR}/${PROJECT_NAME}_export.h
```
jslee02: Could we change the export header to simply `export.h`? I don't see any ambiguity or confliction in including this file if we install this file under `${CMAKE_INSTALL_INCLUDEDIR}/fcl`. In this way, we could include this file as:
```cpp
#include "fcl/export.h" // in FCL code base

#include <fcl/export.h> // in FCL dependent projects.
```

CMakeLists.txt
```diff
@@ -42,6 +45,7 @@ include(FCLMacros)
 include(CompilerSettings)
 include(FCLVersion)
 include(GNUInstallDirs)
+include(GenerateExportHeader)
```
jslee02: I just learned this CMake utility! :open_mouth:

include/fcl/narrowphase/detail/gjk_solver_indep.h
```diff
@@ -50,14 +49,13 @@ namespace detail
...
 template <typename S_>
-struct GJKSolver_indep
+struct FCL_EXPORT GJKSolver_indep
 {
   using S = S_;
 
   /// @brief intersection checking between two shapes
   /// @deprecated use shapeIntersect(const Shape1&, const Transform3<S>&, const Shape2&, const Transform3<S>&, std::vector<ContactPoint<S>>*) const
   template<typename Shape1, typename Shape2>
-  FCL_DEPRECATED
```
jslee02: Is this an oversight or removed intentionally? `fcl_export.h` still has the definition of `FCL_DEPRECATED`. This deprecation was introduced in #52 for backward compatibility. If we want to remove this deprecation, I would like to remove this function entirely not to cause confusion of which `shapeIntersect` should be used.

include/fcl/narrowphase/detail/gjk_solver_libccd.h
```diff
@@ -50,14 +49,13 @@ namespace detail
...
 template <typename S_>
-struct GJKSolver_libccd
+struct FCL_EXPORT GJKSolver_libccd
 {
   using S = S_;
 
   /// @brief intersection checking between two shapes
   /// @deprecated use shapeIntersect(const Shape1&, const Transform3<S>&, const Shape2&, const Transform3<S>&, std::vector<ContactPoint<S>>*) const
   template<typename Shape1, typename Shape2>
-  FCL_DEPRECATED
```
jslee02: ditto:
> Is this an oversight or removed intentionally? fcl_export.h still has the definition of FCL_DEPRECATED. This deprecation was introduced in #52 for backward compatibility. If we want to remove this deprecation, I would like to remove this function entirely not to cause confusion of which shapeIntersect should be used.

include/fcl/narrowphase/detail/traversal/distance/mesh_distance_traversal_node.h
```diff
@@ -212,7 +216,7 @@ bool initialize(
     DistanceResult<S>& result);
 
 template <typename BV>
-FCL_DEPRECATED
+FCL_EXPORT
```
jslee02: I guess this should be `FCL_DEPRECATED_EXPORT`.

j-rivero:
Thanks JS for the review. 

> I guess you replaced "class " with "class FCL_EXPOSE ". It seems this made unintended changes in descriptions of some classes.

You were right. I think I've reverted all the changes you detected. Thanks for carefully looking into them.

>     About the name and installation location of fcl_expose.h.

I agree, we can use `fcl/export.h`. I've modified the code in f98dd43a0ebe981179c0441d9e519b57d55ac4ec to do it.

j-rivero:
While testing DART compilation against this branch I found a little problem with export.h location and fixed it in 831a13e4eb0976f4439ae78c6ec7da4d129cb6dd. In my opinion, we are ready to merge.

jslee02:
> While testing DART compilation against this branch I found a little problem with export.h location and fixed it in 831a13e.

Good catch! Let's merge this unless we have more feedback by tomorrow.

PR #237, opened 2017-11-06
Port of PR #72 to new master

wxmerkt:
This is a partial port of PR #72 by @ricardoglx to the current master. It does not include the [getOctree accessors](https://github.com/flexible-collision-library/fcl/pull/72/commits/13e1b7c2735d11320a405128e7bd69e2a88fb30f) but should be a partial or full fix for #67.

### DIFF
--- a/include/fcl/geometry/bvh/BVH_model.h
+++ b/include/fcl/geometry/bvh/BVH_model.h
@@ -40,7 +40,6 @@
 
 #include <vector>
 #include <memory>
-#include <new>
 
 #include "fcl/math/bv/OBB.h"
 #include "fcl/math/bv/kDOP.h"
--- a/include/fcl/geometry/bvh/BVH_model.h
+++ b/include/fcl/geometry/bvh/BVH_model.h
@@ -40,6 +40,7 @@
 
 #include <vector>
 #include <memory>
+#include <new>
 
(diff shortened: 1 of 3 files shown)


include/fcl/geometry/bvh/BVH_model.h
```diff
@@ -40,6 +40,7 @@
 
 #include <vector>
 #include <memory>
+#include <new>
```
jslee02: Nit: It would be good to move this to `BVH_model-inl.h` as `BVH_model-inl.h` actually uses this header.

wxmerkt:
Thank you very much for the review @jslee02 - I've addressed your comments in 3b2c2e4

jslee02:
[It seems Eigen objects are not supposed to be used with `std::nothrow`](https://travis-ci.org/flexible-collision-library/fcl/jobs/300586958#L1067). Could you revert [this change](https://github.com/flexible-collision-library/fcl/pull/237/commits/3b2c2e4b3f2ca5403a73f776086e7cacc78d7f33#diff-0e6130d1d4ac9b019f5e8c53c4f3aaabR479) (while keeping `#include <new>`)? Sorry for the inconvenience, my bad. :disappointed:

PR #239, opened 2017-11-10
Move make_aligned_shared() to under fcl namespace

jslee02:
Adding definitions under `Eigen` namespace possibly cause confliction with `Eigen` or other libraries. This PR moves `make_aligned_shared()`, which was defined under `Eigen` namespace, to under `fcl` namespace.


jslee02:
I believe the build failures are not due to this change. :thinking: Merging.

PR #240, opened 2017-11-20
Use vcpkg to install dependencies

jslee02:
[`vcpkg`](https://github.com/Microsoft/vcpkg) is a brand new package manager for Windows. I believe it's been mature to installing all the dependencies of FCL. This PR uses `vcpkg` for installing the FCL dependencies (eigen, ccd, octomap) replacing the previous manual installation.

Additionally, this PR changes to test for both VS2015 and VS2017. I'm open to leave this as it was (only for VS2015) or test only for VS2017.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,59 +2,49 @@
 # http://www.appveyor.com/docs/appveyor-yml
 
 version: "{build}"
-os: Visual Studio 2015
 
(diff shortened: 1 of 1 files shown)


jamiesnape:
I can take this on as part of #329.

jslee02:
@jamiesnape It'd be appreciated! Please feel free to close this PR.

Issue #248, opened 2018-01-19
When creating Convex Shape the aabb is always inf.


Levi-Armstrong:
I believe the issue resides [here](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/geometry/shape/convex-inl.h#L101-L102).

Current
```
this->aabb_local.min_.setConstant(-std::numeric_limits<S>::max());
this->aabb_local.max_.setConstant(std::numeric_limits<S>::max());
```

Fixed
```
this->aabb_local.min_.setConstant(std::numeric_limits<S>::max());
this->aabb_local.max_.setConstant(-std::numeric_limits<S>::max());
```

Levi-Armstrong:
I am currently testing a fix and will submit a PR soon.

PR #250, opened 2018-01-24
Convex hull

Levi-Armstrong:
This PR address two items.

- The first was the convex shape's aabb was being initialized incorrectly causing it to always be infinite size.
- The second, is I noticed that when getting distance information for convex shapes the nearest points were appeared to be in the shape frame to the global. After making the changes below they are now in the global frame.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/250)

### DIFF
--- a/include/fcl/geometry/shape/convex-inl.h
+++ b/include/fcl/geometry/shape/convex-inl.h
@@ -50,14 +50,14 @@ class FCL_EXPORT Convex<double>;
 //==============================================================================
 template <typename S>
 Convex<S>::Convex(
-    Vector3<S>* plane_normals, S* plane_dis, int num_planes_,
-    Vector3<S>* points, int num_points_, int* polygons_)
+    Vector3<S>* plane_normals_, S* plane_dis_, int num_planes_,
+    Vector3<S>* points_, int num_points_, int* polygons_)
   : ShapeBase<S>()
 {
-  plane_normals = plane_normals;
-  plane_dis = plane_dis;
+  plane_normals = plane_normals_;
+  plane_dis = plane_dis_;
   num_planes = num_planes_;
-  points = points;
+  points = points_;
   num_points = num_points_;
   polygons = polygons_;
   edges = nullptr;
@@ -98,8 +98,8 @@ Convex<S>::~Convex()
 template <typename S>
 void Convex<S>::computeLocalAABB()
 {
-  this->aabb_local.min_.setConstant(-std::numeric_limits<S>::max());
-  this->aabb_local.max_.setConstant(std::numeric_limits<S>::max());
+  this->aabb_local.min_.setConstant(std::numeric_limits<S>::max());
+  this->aabb_local.max_.setConstant(-std::numeric_limits<S>::max());
   for(int i = 0; i < num_points; ++i)
     this->aabb_local += points[i];
 
--- a/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
+++ b/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
@@ -488,6 +488,7 @@ static int simplexToPolytope2(const void *obj1, const void *obj2,
     // get second support point in opposite direction than supp[0]
     ccdVec3Copy(&dir, &supp[0].v);
     ccdVec3Scale(&dir, -CCD_ONE);
+    ccdVec3Normalize(&dir);
     __ccdSupport(obj1, obj2, &dir, ccd, &supp[1]);
     if (ccdVec3Eq(&a->v, &supp[1].v) || ccdVec3Eq(&b->v, &supp[1].v))
         goto simplexToPolytope2_touching_contact;
@@ -496,6 +497,7 @@ static int simplexToPolytope2(const void *obj1, const void *obj2,
     ccdVec3Sub2(&ab, &supp[0].v, &a->v);
     ccdVec3Sub2(&ac, &supp[1].v, &a->v);
     ccdVec3Cross(&dir, &ab, &ac);
+    ccdVec3Normalize(&dir);
     __ccdSupport(obj1, obj2, &dir, ccd, &supp[2]);
     if (ccdVec3Eq(&a->v, &supp[2].v) || ccdVec3Eq(&b->v, &supp[2].v))
         goto simplexToPolytope2_touching_contact;
@@ -583,6 +585,7 @@ static int simplexToPolytope3(const void *obj1, const void *obj2,
     ccdVec3Sub2(&ab, &b->v, &a->v);
     ccdVec3Sub2(&ac, &c->v, &a->v);
     ccdVec3Cross(&dir, &ab, &ac);
+    ccdVec3Normalize(&dir);
     __ccdSupport(obj1, obj2, &dir, ccd, &d);
     dist = ccdVec3PointTriDist2(&d.v, &a->v, &b->v, &c->v, NULL);
 
@@ -864,6 +867,7 @@ static int nextSupport(const void *obj1, const void *obj2, const ccd_t *ccd,
 {
     ccd_vec3_t *a, *b, *c;
     ccd_real_t dist;
+    ccd_vec3_t dir;
 
     if (el->type == CCD_PT_VERTEX)
         return -1;
@@ -872,14 +876,17 @@ static int nextSupport(const void *obj1, const void *obj2, const ccd_t *ccd,
     if (ccdIsZero(el->dist))
         return -1;
 
-    __ccdSupport(obj1, obj2, &el->witness, ccd, out);
+    ccdVec3Copy(&dir, &el->witness);
+    ccdVec3Normalize(&dir);
+
+    __ccdSupport(obj1, obj2, &dir, ccd, out);
 
     // Compute dist of support point along element witness point direction
     // so we can determine whether we expanded a polytope surrounding the
     // origin a bit.
-    dist = ccdVec3Dot(&out->v, &el->witness);
+    dist = ccdVec3Dot(&out->v, &dir);
 
-    if (dist - el->dist < ccd->epa_tolerance)
+    if ((dist * dist) - el->dist < ccd->epa_tolerance)
         return -1;
 
     if (el->type == CCD_PT_EDGE){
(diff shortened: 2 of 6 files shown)


Levi-Armstrong:
Issue #248

Levi-Armstrong:
Ping?

sherm1:
+@seancurtis-tri do you want to review this?

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/250#-:-LBSPdsM7OzlyHiXqPrc:b-y86c93)*

sherm1:
@levi-armstrong looks like this needs to be merged or rebased to master so it can get through CI.

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/250#-:-LBSPoTzDhNgc2aS_oe3:b-w6t6ib)*

Levi-Armstrong:
@sherm1 Thank you, I just rebased on the master branch.

SeanCurtis-TRI:
I'll review this first thing in the morning (pacific time).

SeanCurtis-TRI:
Several things:

1. You're failing all of CI on this.
2. I've put some comments below about the tolerances used in the tests, looking for strong confirmation that things are as bad as the tests suggest.

Otherwise, this is really nice. Thanks.

---

Reviewed 5 of 5 files at r1.

---

*[include/fcl/narrowphase/detail/gjk_solver_libccd-inl.h, line 565 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/250#-LBY42rH0ayWXln86w8N:-LBY42rH0ayWXln86w8O:b-6ff4ca) ([raw file](https://github.com/flexible-collision-library/fcl/blob/1f508738ee8e20ff6193e98b5b77b2644b7dc269/include/fcl/narrowphase/detail/gjk_solver_libccd-inl.h#L565)):*
> ```objc
> 
>     if (p1)
>       (*p1).noalias() = tf1.inverse(Eigen::Isometry) * *p1;
> ```

FYI Not only is transforming back into the shape's space wrong, but this mathematical operation is wrong. This is *exactly* the usage that [Eigen warns against](https://bitbucket.org/eigen/eigen/src/0b88199f5b4a7cf10f19855dec234fb321220007/Eigen/src/Core/NoAlias.h?at=3.0&fileviewer=file-view-default#NoAlias.h-119).

So, deleting these are good on multiple levels.

---

*[test/test_fcl_signed_distance.cpp, line 71 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/250#-LBY5KCDBPq8Gt0o4Vmf:-LBY5KCDBPq8Gt0o4Vmg:bl45vqv) ([raw file](https://github.com/flexible-collision-library/fcl/blob/5022f9c3061c31af8164e415885a9badb7d44417/test/test_fcl_signed_distance.cpp#L71)):*
> ```cpp
>   res = distance(&s1, tf1, &s2, tf2, request, result);
>   EXPECT_TRUE(res);
>   EXPECT_TRUE(std::abs(result.min_distance - 10) < 1e-6);
> ```

BTW Can you also confirm we can't do any better than 1e-6? I suspect this is a relic.

---

*[test/test_fcl_signed_distance.cpp, line 81 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/250#-LBY-UIEBThyAMKKsIiT:-LBY-UIEBThyAMKKsIiU:bm830zo) ([raw file](https://github.com/flexible-collision-library/fcl/blob/5022f9c3061c31af8164e415885a9badb7d44417/test/test_fcl_signed_distance.cpp#L81)):*
> ```cpp
> 
>   EXPECT_TRUE(res);
>   EXPECT_TRUE(std::abs(result.min_distance - (-5)) < 1e-2);
> ```

BTW This would be better as:

`EXPECT_NEAR(result.min_distance, -5, 1e-2);`

---

*[test/test_fcl_signed_distance.cpp, line 82 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/250#-LBY-aME4uF6XXk8r6H8:-LBY-aME4uF6XXk8r6H9:bswo3yt) ([raw file](https://github.com/flexible-collision-library/fcl/blob/5022f9c3061c31af8164e415885a9badb7d44417/test/test_fcl_signed_distance.cpp#L82)):*
> ```cpp
>   EXPECT_TRUE(res);
>   EXPECT_TRUE(std::abs(result.min_distance - (-5)) < 1e-2);
>   // TODO(JS): The negative distance computation using libccd requires
> ```

This TODO seems to be the result of copy pasta; it's not clear what it's relevance in this test code is.

---

*[test/test_fcl_signed_distance.cpp, line 99 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/250#-LBY05UP8g0FbJZVoChN:-LBY05UP8g0FbJZVoChO:b-dvbssn) ([raw file](https://github.com/flexible-collision-library/fcl/blob/5022f9c3061c31af8164e415885a9badb7d44417/test/test_fcl_signed_distance.cpp#L99)):*
> ```cpp
> 
>   EXPECT_TRUE(res);
>   EXPECT_TRUE(std::abs(result.min_distance - (-1.715728753)) < 1e-2);
> ```

BTW I'd rather see:

```c++
const S expected_distance = tf2.translation().norm() - 30;
EXPECT_NEAR(result.min_distance, expected_distance, 1e-2);
```

That way the "true" answer is always computed to full precision of the scalar S.

---

*[test/test_fcl_signed_distance.cpp, line 101 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/250#-LBY2RN4AQhLYG8UmTmu:-LBY2RN4AQhLYG8UmTmv:b-mb89uz) ([raw file](https://github.com/flexible-collision-library/fcl/blob/5022f9c3061c31af8164e415885a9badb7d44417/test/test_fcl_signed_distance.cpp#L101)):*
> ```cpp
>   EXPECT_TRUE(std::abs(result.min_distance - (-1.715728753)) < 1e-2);
>   // TODO(JS): The negative distance computation using libccd requires
>   // unnecessarily high error tolerance.
> ```

BTW How is it that sphere-sphere distance uses GJK? It should go straight to a special-case function that produces an answer to full machine precision. We certainly have one for determining the distance of separated spheres, but, apparently, not penetrating spheres.  :-P

---

*[test/test_fcl_signed_distance.cpp, line 107 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/250#-LBY0xNYCyBTm2ijeOq0:-LBY0xNYCyBTm2ijeOq1:byk28sd) ([raw file](https://github.com/flexible-collision-library/fcl/blob/5022f9c3061c31af8164e415885a9badb7d44417/test/test_fcl_signed_distance.cpp#L107)):*
> ```cpp
>   if (solver_type == GST_LIBCCD)
>   {
>     EXPECT_TRUE(result.nearest_points[0].isApprox(Vector3<S>(20, 0, 0)));
> ```

BTW `isApprox` is a generally a *terrible* function (see [here](https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#ae8443357b808cd393be1b51974213f9c). If either value ends up close to zero, then the threshold for satisfying the function becomes ridiculously small (almost impossible to satisfy).

This doesn't require an action, per se. I just wanted to inform. You could put a note that because the true answer is far away from zero, this isn't a problem. Ultimately, we need to replace these `isApprox()` calls with something a bit less tempermental.

---


*Comments from [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/250#-:-LBY5T_N7Oj4kN6JFv1f:badkvzq)*

Levi-Armstrong:
@SeanCurtis-TRI I made the changes to the test per requested. To get existing tests to run I had to make some changes but not sure because the distance values are not as accurate as I would have expected. I believe there is something wrong with EPA algorithm but cant seem to figure out what is going on.

SeanCurtis-TRI:
When I run `test_fcl_signed_distance.cpp` I'm getting test failures.

I'm also *really* surprised in the number of normalizations added to make the CI failures go away. Something seems *really* suspicious to me. All you had done up to that point was to leave points in world space instead of object space. It boggles my mind that changes to seemingly unrelated code would need all of this explicit normalization.

Also, I've been doing some testing on the sensitivity of the `gjk_tolerance` in `ColliisionRequest` and have found non-converging behavior as the tolerance gets tighter and tighter. I would assume that `distance` queries with `DistanceRequest` are equally erratic. So, I would mourn, but not be surprised that in penetrating cases, the accuracy goes to hell. This may be an exercise for future PRs. So, for now, we may just have to accept ridiculously bad thresholds.

---

Review status: 4 of 6 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/250#-:-LBblcCCDHWR6UCZeaiP:bnelp7y)*

SeanCurtis-TRI:
Just so you know, I'm currently working on this. I've checked out your PR and have backed up one commit. I really appreciate your work on this, but I know that resolving these issues can be pernicious and a second pair of eyes would help. I'll let you know how it goes tomorrow.

---

Review status: 4 of 6 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/250#-:-LBcFK3O8XZlpLr52iSb:bm0fz31)*

Levi-Armstrong:
I really appreciate you looking into it further. I have for some time now been trying to determine why the FCL EPA implementation results have large errors. I have been using Bullet3 for now for convex distance calculation which does not appear to have this issue. I believe they may have a more advanced implementation of the [EPA algorithm](https://github.com/bulletphysics/bullet3/blob/master/src/BulletCollision/NarrowPhaseCollision/btGjkEpa2.cpp) which may be of value to add to FCL.

Levi-Armstrong:
The main reason for add multiple `ccdVec3Normalize(&dir);` was as I stepped through the code some of the function appeared to assumed the direction vector to be a normalized vector.  Also it appears that the `__ccdSupport` function may assume that the dir vector is normalized. I thought the `extractClosestPoints` was assuming the dir was normalized but I was wrong. I am going through it now and it assumes it is a point in the given simplex. So I will remove those.

[1](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L933-L938)

SeanCurtis-TRI:
This PR is replaced by *two* PRs.

#288 address the CI errors which arose from reporting the nearest points returned by distance in the *world* frame.
#325 Addresses the issues in the `Convex` geometry (including supporting unit tests).

PR #236, opened 2017-10-27
Fixed 'then' and 'else' statements

lunixoid:
I'm a member of the Pinguem.ru competition on finding errors in open source projects. A warning, found using PVS-Studio:
fcl/include/fcl/broadphase/detail/hierarchy_tree-inl.h      1098    err     V523 The 'then' statement is equivalent to the 'else' statement.
fcl/include/fcl/broadphase/detail/hierarchy_tree-inl.h      1116    err     V523 The 'then' statement is equivalent to the 'else' statement.

PR #255, opened 2018-02-14
Cylinder half space bug fix

amcastro-tri:
There is a bug with the penetration query for cylinder vs half space. This is reported in RobotLocomotion/drake#8049.

This PR fixes this bug and implements a new unit test that without the fix would not pass.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/255)

### DIFF
--- a/include/fcl/narrowphase/detail/primitive_shape_algorithm/halfspace-inl.h
+++ b/include/fcl/narrowphase/detail/primitive_shape_algorithm/halfspace-inl.h
@@ -391,7 +391,7 @@ bool cylinderHalfspaceIntersect(const Cylinder<S>& s1, const Transform3<S>& tf1,
   else
   {
     Vector3<S> C = dir_z * cosa - new_s2.n;
-    if(std::abs(cosa + 1) < halfspaceIntersectTolerance<S>() || std::abs(cosa - 1) < halfspaceIntersectTolerance<S>())
+    if(std::abs(cosa) - 1 < halfspaceIntersectTolerance<S>())
       C = Vector3<S>(0, 0, 0);
     else
     {
--- a/include/fcl/narrowphase/detail/primitive_shape_algorithm/halfspace-inl.h
+++ b/include/fcl/narrowphase/detail/primitive_shape_algorithm/halfspace-inl.h
@@ -371,7 +371,7 @@ bool cylinderHalfspaceIntersect(const Cylinder<S>& s1, const Transform3<S>& tf1,
   Vector3<S> dir_z = R.col(2);
   S cosa = dir_z.dot(new_s2.n);
 
-  if(cosa < halfspaceIntersectTolerance<S>())
+  if(std::abs(cosa) < halfspaceIntersectTolerance<S>())
   {
     S signed_dist = new_s2.signedDistance(T);
     S depth = s1.radius - signed_dist;
--- a/test/test_fcl_cylinder_half_space.cpp
+++ b/test/test_fcl_cylinder_half_space.cpp
@@ -1,0 +1,110 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2013-2014, Willow Garage, Inc.
+ *  Copyright (c) 2014-2016, Open Source Robotics Foundation
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of Open Source Robotics Foundation nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/** \author Alejandro Castro */
+
+#include <iostream>
+
+#include "fcl/fcl.h"
+
+#include "gtest/gtest.h"
+
+using namespace std;
+using namespace fcl;
+
+template <typename S>
+void test_collision_cylinder_half_space(fcl::GJKSolverType solver_type)
+{
+  // Numerical precision expected in the results.
+  const double kTolerance = 5 * std::numeric_limits<double>::epsilon();
+
+  const S radius = 0.05;
+  const S length = 4 * radius;
+
+  auto half_space = std::make_shared<Halfspace<S>>(Vector3<S>::UnitZ(), 0.0);
+  auto cylinder = std::make_shared<Cylinder<S>>(radius, length);
+
+  // Pose of cylinder frame C in the world frame W.
+  Transform3<S> X_WC(Translation3<S>(Vector3<S>(0.0, 0.0, 0.049)));
+
+  // Pose of half space frame H in the world frame W.
+  Transform3<S> X_WH = Transform3<S>::Identity();
+
+  CollisionObject<S> half_sapce_co(half_space, X_WH);
+  CollisionObject<S> cylinder_co(cylinder, X_WC);
+
+  fcl::CollisionResult<S> result;
+  static const int num_max_contacts = std::numeric_limits<int>::max();
+  static const bool enable_contact = true;
+  fcl::CollisionRequest<S> request(num_max_contacts, enable_contact);
+  request.gjk_solver_type = solver_type;
+
+  fcl::collide(&half_sapce_co, &cylinder_co, request, result);
+  vector<Contact<S>> contacts;
+  result.getContacts(contacts);
+
+  EXPECT_EQ(static_cast<int>(contacts.size()), 1);
+  EXPECT_NEAR(contacts[0].penetration_depth, 0.051, kTolerance);
+
+  // Now perform the same test but with the cylinder's z axis Cz pointing down.
+  X_WC.linear() = AngleAxisd(M_PI, Vector3d::UnitX()).matrix();
+  X_WC.translation() = Vector3d(0, 0, 0.049);
+  cylinder_co.setTransform(X_WC);
+
+  result.clear();
+  contacts.clear();
+
+  fcl::collide(&half_sapce_co, &cylinder_co, request, result);
+  result.getContacts(contacts);
+
+  EXPECT_EQ(static_cast<int>(contacts.size()), 1);
+  EXPECT_NEAR(contacts[0].penetration_depth, 0.051, kTolerance);
+}
+
+GTEST_TEST(FCL_GEOMETRIC_SHAPES, collision_cylinder_half_space_libccd)
+{
+  test_collision_cylinder_half_space<double>(fcl::GJKSolverType::GST_LIBCCD);
+}
+
+GTEST_TEST(FCL_GEOMETRIC_SHAPES, collision_cylinder_half_space_indep)
+{
+  test_collision_cylinder_half_space<double>(fcl::GJKSolverType::GST_INDEP);
+}
+
+int main(int argc, char* argv[])
+{
+  ::testing::InitGoogleTest(&argc, argv);
+  return RUN_ALL_TESTS();
+}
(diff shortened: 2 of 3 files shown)


amcastro-tri:
@jslee02, would you be able to review this?

test/test_fcl_cylinder_half_space.cpp
```diff
@@ -0,0 +1,110 @@
...
+  fcl::collide(&half_sapce_co, &cylinder_co, request, result);
+  vector<Contact<S>> contacts;
+  result.getContacts(contacts);
+
+  EXPECT_EQ(static_cast<int>(contacts.size()), 1);
+  EXPECT_NEAR(contacts[0].penetration_depth, 0.051, kTolerance);
+
+  // Now perform the same test but with the cylinder's z axis Cz pointing down.
+  X_WC.linear() = AngleAxisd(M_PI, Vector3d::UnitX()).matrix();
+  X_WC.translation() = Vector3d(0, 0, 0.049);
```
jslee02: Nit: It would be good to change `Vector3d` to `Vector3<S>` for completeness.

test/test_fcl_cylinder_half_space.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+  fcl::collide(&half_sapce_co, &cylinder_co, request, result);
+  vector<Contact<S>> contacts;
+  result.getContacts(contacts);
+
+  EXPECT_EQ(static_cast<int>(contacts.size()), 1);
+  EXPECT_NEAR(contacts[0].penetration_depth, 0.051, kTolerance);
+
+  // Now perform the same test but with the cylinder's z axis Cz pointing down.
+  X_WC.linear() = AngleAxisd(M_PI, Vector3d::UnitX()).matrix();
```
jslee02: Nit: It would be good to use AngleAxis<S> for completeness.

amcastro-tri:
I saw the cone problem also. I don't think I'll have time to push a unit test any time soon. I'll see to write a unit test for that one on Monday.

@jslee02, is there a reason to use `halfspaceIntersectTolerance<S>()` in the comparisons? I don't see why not to compare against `0`. `halfspaceIntersectTolerance<S>()` resolves to [1e-7 here](https://github.com/flexible-collision-library/fcl/blob/master/src/narrowphase/detail/primitive_shape_algorithm/halfspace.cpp#L55).

test/test_fcl_cylinder_half_space.cpp
amcastro-tri: Excellent, that's right. I was just copying the patter in other unit tests and I templated everything on `<S>`. Did you have a chance to test FCL with `AutoDiffXd`? the automatic differentiation scalar from Eigen.

amcastro-tri: I wonder what it would take to be able to say: `test_collision_cylinder_half_space<AutoDiffXd>(fcl::GJKSolverType::GST_INDEP);`. Here in the development of [Drake](https://github.com/RobotLocomotion/drake) at [TRI](http://www.tri.global/) we would love to be able to do just that! :-)

jslee02: I wrote a very simple test case with `AutoDiffScalar` [here](https://github.com/flexible-collision-library/fcl/blob/a91b931e9824813b31474e721b86d52fb11ba1b9/test/test_fcl_auto_diff.cpp). Is that what you're looking for?

jslee02:
> I saw the cone problem also. I don't think I'll have time to push a unit test any time soon. I'll see to write a unit test for that one on Monday.

No worries. Let me create an issue for this to track.

> is there a reason to use halfspaceIntersectTolerance<S>() in the comparisons? I don't see why not to compare against 0. halfspaceIntersectTolerance<S>() resolves to 1e-7 here.

I didn't write the code, but here is my hunch. Zero tolerance would be too strict criteria for an object (cylinder or cone) to be regarded laid on the ground parallel to the ground surface. Suppose you drop an object on the ground (halfspace) in a physics simulator, then the object would jitter a lot before it becomes stable due to the on-and-off contact points on the end of edges of the side (e.g., the contact points on the top and bottom of the cylinder).

amcastro-tri:
> Suppose you drop an object on the ground (halfspace) in a physics simulator, then the object would jitter a lot before it becomes stable due to the on-and-off contact points on the end of edges of the side (e.g., the contact points on the top and bottom of the cylinder).

That is a problem of the simulator what to do with the information you provide. The problem you mention is because real physical contact does not happen at points but on surfaces patches. I believe that as a geometry engine, the penetration queries should be purely geometrical, irrespective of what a physics engine might do with the results. Attempting to couple them with tricks as using a loose tolerance will most likely bite you back and produce difficult to find bugs. 

I think the right thing to do would be doing the check against "zero", though I'd need to investigate the math in this particular case further to make sure there is no other pathological case.

jslee02:
> That is a problem of the simulator what to do with the information you provide. The problem you mention is because real physical contact does not happen at points but on surfaces patches. I believe that as a geometry engine, the penetration queries should be purely geometrical, irrespective of what a physics engine might do with the results. Attempting to couple them with tricks as using a loose tolerance will most likely bite you back and produce difficult to find bugs.

That totally makes sense. Thinking further, I actually presume that's because "comparing floating point number to zero is not safe". Probably, we never get to the if-statement is the true case if we compare `cosa` to zero. There are many ways of comparing a floating point number to zero, and here we use one of the way (i.e., using fixed threshold).

I'm open to using a better way for this.

amcastro-tri:
Yes, on further thought I agree with you. Using a tolerance is a good idea. However, what is the best way to choose this tolerance? shouldn't it be a  number closer to machine epsilon? 1e-7 seems a bit loose to me but there might be a good reason for it.

test/test_fcl_cylinder_half_space.cpp
amcastro-tri: Excellent! is `fcl::collide()` also supported? will it work with meshes?

jslee02: I don't see any particular reason `fcl::collide()` wouldn't work as long as everything is templated for the scalar type, which I believe so, but I haven't tested it myself.

jslee02:
> what is the best way to choose this tolerance?

Sorry, it's a tricky question to answer to me. :sweat:

sherm1:
Travis CI is failing for this PR. Is that due to this change?

SeanCurtis-TRI:
I also need to ponder this.

@amcastro-tri makes a compelling point about the geometry query introducing error before the physics gets a chance to reason at all.

However, the idea of testing a float against zero feels categorically wrong. For an arbitrarily oriented plane, it seems that the odds that a cylinder would ever be considered as "lying" on the plane would be roughly 1 in 2 *billion*. It's like testing for vertex-vertex collision -- simply not worth it.

It comes down to the fact that the geometric query itself is subject to numerical accuracy and it should do its best effort to come up with the right geometric answer which accounts for numerical. I think operating with respect to an epsilon in this regard is inevitable. It should be well reasoned and it may be that 1e-7 is far too large for doubles.

But, at the end of the day, strictly comparing against 0 just seems suspicious.

test/test_fcl_cylinder_half_space.cpp
```diff
@@ -0,0 +1,110 @@
...
+  auto half_space = std::make_shared<Halfspace<S>>(Vector3<S>::UnitZ(), 0.0);
+  auto cylinder = std::make_shared<Cylinder<S>>(radius, length);
+
+  // Pose of cylinder frame C in the world frame W.
+  Transform3<S> X_WC(Translation3<S>(Vector3<S>(0.0, 0.0, 0.049)));
+
+  // Pose of half space frame H in the world frame W.
+  Transform3<S> X_WH = Transform3<S>::Identity();
+
+  CollisionObject<S> half_sapce_co(half_space, X_WH);
```
sherm1: sapce -> space

sherm1:
One rational way to choose tolerances is to view them as the result of an optimization in which we try to make a numerical calculation as close to the analytical result as possible. A tolerance of 0 won't do that because the computation will become dominated by roundoff error. A tolerance of 0.1 won't do that because it is a bad approximation of the geometry (sometimes called "truncation error"). Somewhere in between is a tolerance that minimizes some measure of the actual error.

In the case of this cylinder vs half space algorithm, I believe the tolerance is being used to choose one of three discrete contact point possibilities: center of an end cap, edge of an end cap, or center of a long edge of the tube. If the objects couldn't interpenetrate, the contact points really would change discretely. But these actually interpenetrate, so there is an overlap volume that evolves continuously. That provides a "right answer" that we could hope to approximate: that the contact "point" is at the centroid of the overlap volume. Ideally the algorithm would just return that point! But if it has to instead make discrete choices, the ideal tolerance would be the one that minimizes a measure of the deviation between the returned point and that centroid. Of course it is easier said than done to make that determination, but at least there is a reasonable way to think about it.

SeanCurtis-TRI:
I've retriggered CI - one test was clearly a flake (and has since passed). The two mac tests seem to fail on an unrelated test. I've retriggered them and may take a peek later if they still fail.

include/fcl/narrowphase/detail/primitive_shape_algorithm/halfspace-inl.h
```diff
@@ -371,7 +371,7 @@ bool cylinderHalfspaceIntersect(const Cylinder<S>& s1, const Transform3<S>& tf1,
   Vector3<S> dir_z = R.col(2);
   S cosa = dir_z.dot(new_s2.n);
 
-  if(cosa < halfspaceIntersectTolerance<S>())
+  if(std::abs(cosa) < halfspaceIntersectTolerance<S>())
```
SeanCurtis-TRI: I'm putting this comment here because github won't let me put it where I *want* to.

On line 394 we have the line of code:

```C++
if (std::abs(cosa + 1) < halfspaceIntersectTolerance<S>() || std::abs(cosa - 1) < halfspaceIntersectTolerance<S>())
```

It could be replaced with:

```C++
if (std::abs(cosa) - 1 < halfspaceIntersectTolerance<S>())
```

I figured, as long as we're in this code, we could patch this as well (although this isn't a correctness issue, merely an efficiency/compactness issue).

SeanCurtis-TRI:
@jslee02 The mac CI is failing on (what seems to be) an unrelated test. I don't have access to a mac; any idea what the best way to resolve this is?

---

Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/flexible-collision-library/fcl/255#-:-L5X3bNjXgkfCt3O_rYx:bdabexw)*

amcastro-tri:
@jslee02 the mac CI keeps failing and it doesn't seem to be an obvious problem. Should we create an issue for this? do you think you'd have an idea on what causes the problem?

SeanCurtis-TRI:
I added issue #260.

jslee02:
@amcastro-tri I retriggered Travis for master branch builds, which were failed due to failing in installing `homebrew-science/libccd`, to see if the current failing is actually due to this change. Since `libccd` is now available at homebrew-core, the master branch builds should be able to install `libccd`.

jslee02:
Okay, it seems the failure was introduced by #236 because Build #612 passed but not Build #626 (after #236 merged).

SeanCurtis-TRI:
I went ahead and addressed the comments.

---

Reviewed 2 of 3 files at r1, 1 of 1 files at r2, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[include/fcl/narrowphase/detail/primitive_shape_algorithm/halfspace-inl.h, line 374 at r2](https://reviewable.io:443/reviews/flexible-collision-library/fcl/255#-L5X3OmdmjPKRJFB9RQI-r2-374:-L6bdXvRkZQPijYErGcD:bb6un1j) ([raw file](https://github.com/flexible-collision-library/fcl/blob/0d6600a56757d0f2b5fb256776c2a1a4c2cf586e/include/fcl/narrowphase/detail/primitive_shape_algorithm/halfspace-inl.h#L374)):*

I'm putting this comment here because github won't let me put it where I *want* to.

On line 394 we have the line of code:

```C++
if (std::abs(cosa + 1) < halfspaceIntersectTolerance<S>() || std::abs(cosa - 1) < halfspaceIntersectTolerance<S>())
```

It could be replaced with:

```C++
if (std::abs(cosa) - 1 < halfspaceIntersectTolerance<S>())
```

I figured, as long as we're in this code, we could patch this as well (although this isn't a correctness issue, merely an efficiency/compactness issue).
</blockquote></details>

Done

---

*[test/test_fcl_cylinder_half_space.cpp, line 82 at r1](https://reviewable.io:443/reviews/flexible-collision-library/fcl/255#-L5X3OmdmjPKRJFB9RQK-r1-82:-L6bd_1dtLjiMYMbOI5b:bb6un1j) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4a5152d4d2ad311fbf508eea116faeea55e6e286/test/test_fcl_cylinder_half_space.cpp#L82)):*
<details><summary><i>Previously, jslee02 (Jeongseok Lee) wrote…</i></summary><blockquote>

Nit: It would be good to use `AngleAxis<S>` for completeness.
</blockquote></details>

Done

---

*[test/test_fcl_cylinder_half_space.cpp, line 83 at r1](https://reviewable.io:443/reviews/flexible-collision-library/fcl/255#-L5X3OmdmjPKRJFB9RQK-r1-83:-L6bdd9vhDPDNCTr7Ml3:bb6un1j) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4a5152d4d2ad311fbf508eea116faeea55e6e286/test/test_fcl_cylinder_half_space.cpp#L83)):*
<details><summary><i>Previously, jslee02 (Jeongseok Lee) wrote…</i></summary><blockquote>

I don't see any particular reason `fcl::collide()` wouldn't work as long as everything is templated for the scalar type, which I believe so, but I haven't tested it myself.
</blockquote></details>

Done

---

*[test/test_fcl_cylinder_half_space.cpp, line 65 at r2](https://reviewable.io:443/reviews/flexible-collision-library/fcl/255#-L5X3OmdmjPKRJFB9RQK-r2-65:-L6bdZICYYOoIL8b0oo1:bb6un1j) ([raw file](https://github.com/flexible-collision-library/fcl/blob/0d6600a56757d0f2b5fb256776c2a1a4c2cf586e/test/test_fcl_cylinder_half_space.cpp#L65)):*

sapce -> space
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/flexible-collision-library/fcl/255#-:-L6brPxzH0FpXBGFyJkJ:bm89j0m)*

amcastro-tri:
Thank you  @SeanCurtis-TRI! :lgtm: 

---

Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/flexible-collision-library/fcl/255#-:-L6c91XrCxpD8utI6_Fj:b96cb5q)*

sherm1:
:lgtm:

---

Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/flexible-collision-library/fcl/255#-:-L6dIbKoEvNq1yy7XD3B:bnfp4nl)*

PR #266, opened 2018-03-09
Move travis CI to use xcode 9 instead of 7.3

SeanCurtis-TRI:
Various mac errors seem to go away by advancing from xcode 7.3 to later versions. This tests it under travis.ci.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/266)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,11 +20,11 @@ matrix:
       compiler: clang
       env: BUILD_TYPE=Release COVERALLS=OFF
     - os: osx
-      osx_image: xcode7.3
+      osx_image: xcode9
       compiler: clang
       env: BUILD_TYPE=Debug COVERALLS=OFF
     - os: osx
-      osx_image: xcode7.3
+      osx_image: xcode9
       compiler: clang
       env: BUILD_TYPE=Release COVERALLS=OFF
 


SeanCurtis-TRI:
This may or may not merge. In trying to alleviate commit problems, I found that AppleClang 9 is working great. So, I'm trying a sampling to see if TravisCI behaves similarly.

SeanCurtis-TRI:
Moving to xcode 9 didn't fix the problems with the `box_capusle_1` test. However, it did resolve the most recent issues with `fcl_test_box_box.cc`. 

Inexplicably, I'm running this on another mac with (what appears to be) the same configuration and the failing tests *don't* fail. I'm unsure why it continues to fail in CI.

@jslee02 Let me know if you still approve jumping all the way to XCode 9.

jslee02:
> Let me know if you still approve jumping all the way to XCode 9.

I'm on board using the newer version of macOS.

However, if some tests fail on the old version of macOS + Xcode (< Xcode 9 + OS X 10.12), it could be an issue if someone wants to use FCL in that configuration. I'm not sure if that case is considerable, but I think we could revisit this at that time.

SeanCurtis-TRI:
I agree. What is the oldest version of XCode fcl wants to support?

jslee02:
I don't use macOS actively so it would be better someone else suggests that. :sweat_smile:

scpeters:
Homebrew supports and builds binaries (bottles) for the 3 most recent releases (10.11 el capitan, 10.12 sierra, 10.13 high sierra). It's a big pain to support anything older than that. For CI, we just use test on one distro for gazebo, though we make sure that things compile and build on the 3 versions that homebrew supports.

scpeters:
if you set an environment variable `CTEST_OUTPUT_ON_FAILURE=1`, then it will show the output of the failing tests

SeanCurtis-TRI:
The problem I'm having with CI is *not* seeing the error details. It is reproducing the errors anywhere outside of travis. Travis is currently configured to report the error logs. But, try as I might, I can't get failures on the mac I have access to. (Hence the iterative changes in this PR with the hope that bringing CI closer to "my" mac, the errors would go away. Such has *not* proven to be the case.) My problem is that before Friday, I had never touched a mac in my life. So, I know nothing about xcode, bottles, homebrew, or how to interpret osx versions. I'm putting this on the back burner for now..it's clear this is going to be a big, uphill battle for me and I don't have the cycles to spare.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/flexible-collision-library/fcl/266#-:-L7PbRnCZkHfi-cM3Vbw:b-n56t5n)*

jslee02:
> Homebrew supports and builds binaries (bottles) for the 3 most recent releases (10.11 el capitan, 10.12 sierra, 10.13 high sierra).

How about we support the latest image of each releases? According to [this doc](https://docs.travis-ci.com/user/reference/osx/), they are:
* xcode8 (10.11)
* xcode9.2 (10.12)
* xcode9.4 (10.13)

Also, if we build with debug and release mode for each image, the build matrix size would be too large. So I suggest building with both of debug and release only for the latest (xcode9.4) and building with release for others.

By the way, I don't have a strong preference on this. So I'm totally fine with merging this PR as it is.

PR #270, opened 2018-03-26
Gitmask Anonymous PR

gitmask-anonymous:
This is an anonymous PR submitted via Gitmask - https://www.gitmask.com

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/270)

### DIFF
--- a/include/fcl/math/motion/interp_motion-inl.h
+++ b/include/fcl/math/motion/interp_motion-inl.h
@@ -58,6 +58,8 @@ InterpMotion<S>::InterpMotion()
   // Default reference point is local zero point
 
   // Default linear velocity is zero
+  linear_vel = 0;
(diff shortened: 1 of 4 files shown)


sherm1:
Thanks for your contribution! Totally respect your desire for anonymity, but that should be limited to your identity, not the content of the PR. Please edit the PR description to say what this change is for, and the commit should also have a relevant description so that the repo history will make sense if someone needs to look at it later. There should also be a unit test verifying that this change works, and ensuring that it won't get broken later by some programmer who doesn't understand it.

sherm1:
Per discussion with anonymous OP, closing this GitMask PR in favor of new anonymous GitHub user @geoynomous (PR #271)

PR #288, opened 2018-05-08
Correct distance queries to report nearest point in world frame

SeanCurtis-TRI:
Changed *slightly* from the original PR submission.

Now this focuses on correcting the return values in the distance queries. `DistanceResult` reports that the near points are reported in the *world* frame. However, this was not supported by the code. This PR changes the various distance queries to report world-frame values and updates the tests to reflect this.

~~This PR is an alternate variation of #250 -- trying to get through CI.~~

- ~~The first was the convex shape's aabb was being initialized incorrectly causing it to always be infinite size.~~
- ~~The second, is I noticed that when getting distance information for convex shapes the nearest points were appeared to be in the shape frame to the global. After making the changes below they are now in the global frame.~~

~~NOTE: Part of this PR introduces a signed-distance test between a sphere and capsule. The test revealed an error in the calculation. The test documents this error.~~

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/288)

### DIFF
--- a/include/fcl/narrowphase/detail/gjk_solver_indep-inl.h
+++ b/include/fcl/narrowphase/detail/gjk_solver_indep-inl.h
@@ -625,8 +625,10 @@ struct ShapeDistanceIndepImpl
 
       if(distance) *distance = (w0 - w1).norm();
 
-      if(p1) *p1 = w0;
-      if(p2) (*p2).noalias() = shape.toshape0.inverse() * w1;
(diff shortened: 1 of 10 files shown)


Levi-Armstrong:
@SeanCurtis-TRI I built everything using this PR along with #325  and all of my test now pass. :+1:

Levi-Armstrong:
@SeanCurtis-TRI Are you wanting me to review this Reviewable?

Issue #293, opened 2018-05-28
Nan value in the nearest points when using distance()


tongtybj:
While doing distance check for two boxes, I found ```nan``` values in the nearest points after using distance().
One of the sample is as follows, and two boxes do not collide.

```
   DistanceResult<double> result;
   result.clear();
   DistanceRequest<double> request(true);

   std::shared_ptr< CollisionGeometry<double> > cgeomBox1(new Box<double>(2.750000, 6.000000, 0.050000));
   std::shared_ptr< CollisionGeometry<double> > cgeomBox2(new Box<double>(0.424000, 0.150000, 0.168600));
   CollisionObject<double> objBox1(cgeomBox1, Eigen::Quaterniond(1,0,0,0).matrix(), Eigen::Vector3d(1.625000, 0.000000, 0.500000));
   CollisionObject<double> objBox2(cgeomBox2, Eigen::Quaterniond(0.672811, 0.340674, 0.155066, 0.638138).matrix(), Eigen::Vector3d(0.192074, -0.277870, 0.273546));

   fcl::distance(&objBox1, &objBox2, request, result);
   std::cout << "distance = " << result.min_distance << std::endl;
   std::cout << " point on obj1: x = " << result.nearest_points[0](0) << " y = " << result.nearest_points[0](1) << " z = " << result.nearest_points[0](2) << std::endl;
   std::cout << " point on obj2: x = " << result.nearest_points[1](0) << " y = " << result.nearest_points[1](1) << " z = " << result.nearest_points[1](2) << std::endl;
```

Can someone help me to figure out the reason, or give me some hint about the algorithm of distance calculation between two boxes?

SeanCurtis-TRI:
Thanks for the detailed reproduction info. I'll look into this and get back to you.

tongtybj:
@SeanCurtis-TRI 
Thank you for your response.

What I have found is that the [normal  ```n```](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L1124-L1125) is zero, whiling processing the function ```extractClosestPoints```, and this is caused since vector [```AB``` and ```AC```](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L1090-L1094) are parallel.  Right now, I am studying about the algorithm to figure out why triangle ```ABC```  is corrupt.

PR #295, opened 2018-05-29
Additional process to extract closest points in convexities when simplex is not triangle

tongtybj:
This PR is associated with #293.

 I found sometimes the simplex may shrink from triangle to line segment, while calculating the min distance and closest points in convexities based on GJK. The box-box sample in #293 is one of the case.
 
So in this commit, I simply copy the code  from [``` else if (simplex_size ==2) ...```](https://github.com/tongtybj/fcl/blob/master/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L1021-L1082), if the simplex is judged as line segment.

I have confirmed this commit works with the case of #293. However, I am not very sure whether this change is theoretically correct, I am new to the GJK algorithm.

I hope this rough propose can be helpful.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/295)

### DIFF
--- a/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
+++ b/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
@@ -1093,61 +1093,153 @@ static void extractClosestPoints(ccd_simplex_t* simplex,
     ccdVec3Sub2(&AC, &(simplex->ps[2].v), &(simplex->ps[0].v));
     ccdVec3Cross(&n, &AB, &AC);
 
-    // Since p lies in ABC, it can be expressed as
-    //
-    // p = A + p'
-    // p' = s*AB + t*AC
-    //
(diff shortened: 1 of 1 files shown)


hongkai-dai:
Thanks @tongtybj, I believe @SeanCurtis-TRI has been working on this.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/295#-:-LDiJj4G1WijKbADdJc9:b-lmtg5b)*

SeanCurtis-TRI:
Yep, thanks @tongtybj. You've basically got the right idea, but we don't quite want to go with this solution. I'm working up a more general solution this morning. We definitely want to avoid copy-and-pasting code.  :) Let's put this on hold while I get an alternative put together.

SeanCurtis-TRI:
Closed in favor of #296.

PR #300, opened 2018-06-03
add missing template instantiation for conservativeAdvancement

### LINKED ISSUE #297: Missing template instantiation of `conservativeAdvancement` for RSS/OBBRSS
It uses the instantiation of with `const BVHModel<BV>&`,

but then this is used as node 

  `MeshConservativeAdvancementTraversalNode<BV> node;`

and not

  `MeshConservativeAdvancementTraversalNodeRSS<BV> node;` or
   `MeshConservativeAdvancementTraversalNodeOBBRSS<BV> node;`

The two later one works thread-safe while the previous one does not as its `initialize` refits the tree.

geoynomous:
Now partial template instantiation are available for RSS/OBBRSS
which resemble the situation of 0.5.0 where

MeshConservativeAdvancementTraversalNodeRSS/OBBRSS are used and not
MeshConservativeAdvancementTraversalNode

The later refitting the whole tree, while the former two only work with
transformations locally stored in the node.

Refactors code to avoid duplication

resolves #297

Couldn't find a test to adap

### DIFF
--- a/include/fcl/narrowphase/detail/conservative_advancement_func_matrix-inl.h
+++ b/include/fcl/narrowphase/detail/conservative_advancement_func_matrix-inl.h
@@ -70,17 +70,16 @@ namespace fcl
 namespace detail
 {
 
-//==============================================================================
-template<typename BV>
-bool conservativeAdvancement(const BVHModel<BV>& o1,
-                             const MotionBase<typename BV::S>* motion1,
-                             const BVHModel<BV>& o2,
-                             const MotionBase<typename BV::S>* motion2,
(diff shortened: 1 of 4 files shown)


geoynomous:
I'll check with other compilers ....

geoynomous:
It seems it's a VS2015 problem .... can somebody check with a newer version of VS?

geoynomous:
Or can somebody add a workaround on top of my commit? Fact is that the current `master` does **not** use the correct derived `MeshConservativeAdvancementTraversalNodeRSS/OBBRSS` classes.

PR #302, opened 2018-06-04
Fix the bug in penEPAPosClosest

hongkai-dai:
FCL extracts the deepest penetration points `p1, p2` in the function `penEPAPosClosest`. The implementation within this function is wrong. It loops through all vertices of the polytope from EPA algorithm, and takes the vertex that is closest to the origin as the point `p1-p2`. But `p1-p2` should correspond to the point on the boundary of the polytope, that is nearest to the origin, not on the vertices. As a matter of fact, that nearest point on the polytope's boundary is computed through EPA algorithm already, so we only need to construct the simplex as the active feature of that nearest point, and then we can easily extract `p1, p2` from the simplex, same as we did in GJK algorithm.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/302)

### DIFF
--- a/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
+++ b/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
@@ -1249,6 +1249,17 @@ static inline ccd_real_t _ccdDist(const void *obj1, const void *obj2,
   return -CCD_REAL(1.);
 }
 
+/**
+ * Given the nearest point on the polytope inside the Minkowski sum A⊖B, returns
+ * the point p1 on geometric object A and p2 on geometric object B, such that
+ * p1 is the deepest penetration point on A, and p2 is the deepest penetration
+ * point on B.
+ * @param[in] pt The polytope inside Minkowski sum A⊖B. Unused in this function.
+ * @param[in] nearest The point that is nearest to the origin on the boundary of
+ * the polytope.
+ * @param[out] p1 the deepest penetration point on A.
+ * @param[out] p2 the deepest penetration point on B.
+ */
 static int penEPAPosClosest(const ccd_pt_t *pt, const ccd_pt_el_t *nearest,
                             ccd_vec3_t *p1, ccd_vec3_t* p2)
 {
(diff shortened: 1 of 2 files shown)


sherm1:
Needs a merge with master.
Review pass done, ptal.

---

---

*[include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h, line 1263 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/302#-LEAgIMg80A4zwITHD_M:-LEAgIMg80A4zwITHD_N:bo53zpb) ([raw file](https://github.com/flexible-collision-library/fcl/blob/efcb3ac545fee191971afaf4aa7a26a5867ab29c/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L1263)):*
> ```objc
>  * @param[out] p2 the deepest penetration point on B.
>  */
> static int penEPAPosClosest(const ccd_pt_t *pt, const ccd_pt_el_t *nearest,
> ```

Return value is undocumented.

---

*[include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h, line 1264 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/302#-LEAfmvEDhjrknWopsmM:-LEAfmvEDhjrknWopsmN:b-aavvzx) ([raw file](https://github.com/flexible-collision-library/fcl/blob/efcb3ac545fee191971afaf4aa7a26a5867ab29c/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L1264)):*
> ```objc
>  */
> static int penEPAPosClosest(const ccd_pt_t *pt, const ccd_pt_el_t *nearest,
>                             ccd_vec3_t *p1, ccd_vec3_t* p2)
> ```

Nit - inconsistent placement of `*`. Please put next to the type as for p2.

---

*[include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h, line 1276 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/302#-LEAgYhDEAPrTG05cneY:-LEAgYhDEAPrTG05cneZ:bdpszxz) ([raw file](https://github.com/flexible-collision-library/fcl/blob/efcb3ac545fee191971afaf4aa7a26a5867ab29c/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L1276)):*
> ```objc
>     ccdVec3Copy(p2, &v->v.v2);
>     return 0;
>   } else {
> ```

Since the above case returns unconditionally, it is better not to use an else here. The code is more readable without it since it loses one level of indentation.

---

*[include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h, line 1332 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/302#-LEAhlDC0e1_mBTZKZWR:-LEAhlDC0e1_mBTZKZWS:bh96kb7) ([raw file](https://github.com/flexible-collision-library/fcl/blob/efcb3ac545fee191971afaf4aa7a26a5867ab29c/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L1332)):*
> ```objc
>       throw std::logic_error(
>           "Unsupported point type. The closest point should be either a "
>           "vertex, on an edge, or an a face.\n");
> ```

Minor: an->on

---

*[include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h, line 1334 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/302#-LEAiCXr1RCER_7yUBEy:-LEAiCXr1RCER_7yUBEz:bsoumo4) ([raw file](https://github.com/flexible-collision-library/fcl/blob/efcb3ac545fee191971afaf4aa7a26a5867ab29c/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L1334)):*
> ```objc
>           "vertex, on an edge, or an a face.\n");
>     }
>     // Now compute the closest point in the simplex.
> ```

I don't understand this comment. Closest to what? And I think it is finding two points?

---

*[include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h, line 1345 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/302#-LEAikEaBO2EBYo5XoDy:-LEAikEaBO2EBYo5XoDz:b-fjcc39) ([raw file](https://github.com/flexible-collision-library/fcl/blob/efcb3ac545fee191971afaf4aa7a26a5867ab29c/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L1345)):*
> ```objc
>     return 0;
>   }
>   FCL_UNUSED(pt);
> ```

FYI probably better at top of method?

---

*[test/narrowphase/detail/convexity_based_algorithm/test_gjk_libccd-inl.cpp, line 230 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/302#-LEAekhO1VYRwJmViBEI:-LEAekhO1VYRwJmViBEJ:b-ai3833) ([raw file](https://github.com/flexible-collision-library/fcl/blob/efcb3ac545fee191971afaf4aa7a26a5867ab29c/test/narrowphase/detail/convexity_based_algorithm/test_gjk_libccd-inl.cpp#L230)):*
> ```cpp
>     const double expected_distance = -X_WB2.translation()(0) - 1;
>     CheckDistance(X_WB2, expected_distance, Vector2<S>(-0.5, X_WB2.translation()(1)), Vector2<S>(X_WB2.translation()(0) + 0.5, X_WB2.translation()(1)), tol);
>   };
> ```

Nit - line lengths above.

---


*Comments from [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/302#-:-LEAdMPnD3Al83BN1t8W:b-g9s7gn)*

hongkai-dai:
This PR reveals more bugs in EPA implementation. I will close this PR now, fix the bugs, and then submit it later.

PR #308, opened 2018-06-20
Add GenerateBVHSubModel variants

nicovanduijn:
Identical to the GenerateBVHModel, but without ending the model, such that
more geometric primitives can be added. This is useful if you want to
create a BVHModel containing multiple geometric primitives.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/308)


sherm1:
This looks useful. I took a look and made some comments. However, it would be good for someone else to look at this to verify there are no unexpected consequences.

Currently all changes to FCL should be made very cautiously because there are very few unit tests ensuring that the functionality is as-specified. It's very important to include unit tests with all new code so that we don't compound this problem.

---

Reviewed 5 of 5 files at r1.
Review status: all files reviewed, 4 unresolved discussions (waiting on @nicovanduijn)

---

*[include/fcl/geometry/geometric_shape_to_BVH_model.h, line 58 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/308#-LGXDgvB53UTQDfrFB1a:-LGXDgvB53UTQDfrFB1b:b-ma8g40) ([raw file](https://github.com/flexible-collision-library/fcl/blob/fcb99cfc452085407eeb6e2d1c0fcb1ed2a82303/include/fcl/geometry/geometric_shape_to_BVH_model.h#L58)):*
> ```objc
> /// @brief Generate BVH model from box
> template<typename BV>
> void generateBVHModel(BVHModel<BV>& model, const Box<typename BV::S>& shape, const Transform3<typename BV::S>& pose, bool submodel = false);
> ```

The meaning of the name `submodel` seems unclear to me. Consider calling this something that captures its functionality better, like `dont_end_model_yet` or `more_shapes_to_come` or something like that.

Also, the new parameter needs to be documented in the Doxygen docs for each API method (or you could put them in a Doxygen group and provide group documentation).

---

*[test/test_fcl_broadphase_distance.cpp, line 266 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/308#-LGXEYec00MF8nnmwlAh:-LGXEYec00MF8nnmwlAi:bwb2ahg) ([raw file](https://github.com/flexible-collision-library/fcl/blob/fcb99cfc452085407eeb6e2d1c0fcb1ed2a82303/test/test_fcl_broadphase_distance.cpp#L266)):*
> ```cpp
>     Sphere<S> sphere(single_size / 2);
>     BVHModel<OBBRSS<S>>* model = new BVHModel<OBBRSS<S>>();
>     generateBVHModel(*model, sphere, Transform3<S>::Identity(), 16, 16, false);
> ```

Since the original behavior is the default you don't need to add these repetitive `false` parameters to all these test. You can just leave them as they were before so the changes don't need to be reviewed and so this PR has fewer changes.

On the other hand, I didn't see a unit test of the new functionality anywhere. Please add tests that show proper functioning of the new code that will ensure the code continues to function as you intended when other programmers make changes later, perhaps not understanding your intent.

---

*[test/test_fcl_shape_mesh_consistency.cpp, line 67 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/308#-LGXFYqHAuit236b4djY:-LGXFYqHAuit236b4djZ:b-n9d1ep) ([raw file](https://github.com/flexible-collision-library/fcl/blob/fcb99cfc452085407eeb6e2d1c0fcb1ed2a82303/test/test_fcl_shape_mesh_consistency.cpp#L67)):*
> ```cpp
> 
>   generateBVHModel(s1_rss, s1, Transform3<S>::Identity(), 16, 16, false);
>   generateBVHModel(s2_rss, s2, Transform3<S>::Identity(), 16, 16, false);
> ```

Ditto -- no need for these `false` arguments, but a definite need for some `true` arguments!

---

*[test/test_fcl_utility.h, line 597 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/308#-LGXFi_AA_RXBiQFLfaR:-LGXFi_AA_RXBiQFLfaS:b-jrp16a) ([raw file](https://github.com/flexible-collision-library/fcl/blob/fcb99cfc452085407eeb6e2d1c0fcb1ed2a82303/test/test_fcl_utility.h#L597)):*
> ```objc
>   {
>     BVHModel<OBBRSS<S>>* model = new BVHModel<OBBRSS<S>>();
>     generateBVHModel(*model, sphere, Transform3<S>::Identity(), 16, 16, false);
> ```

Ditto: no need for `false` here. Do you need a `true` test here as well?

---


*Comments from [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/308#-:-LGXG0DpDQ9HHVhirPMF:b-guiucf)*

nicovanduijn:
Thanks for the review @sherm1 

Regarding your comments:
- Indeed my goal was to break as little pre-existing code as possible, thus I made the boolean flag optional. However, the calls
```
    generateBVHModel(*model, sphere, Transform3<S>::Identity(), 16, 16, false);
```
Definitely needed the "false" flag because otherwise the compiler doesn't know whether to cast the second "16" to boolean and call the overload with num_faces instead. This might be solved by replacing 16 with 16U everywhere, but might not be the safest either.

The best I could come up with that is safe, breaks little existing code and avoids lots of code duplication was introducing an enum class to replace the boolean.

I added comments and some unit tests. I'm not very experienced in writing unit tests, so let me know if this is appropriate.

Any reason the tests `test_fcl_capsule_box_2` fails on my machine, while it works on CI? And the other way around for one of the other unit tests which I never touched either.

SeanCurtis-TRI:
Nico, thanks for the PR. Sorry to make the experience a bit like pulling teeth and I'm particularly grateful for how willing you were to play along.

PR #316, opened 2018-07-24
Add primitive box-sphere collision and distance query

SeanCurtis-TRI:
By default, the GJK solver was being used for performing distance queries between box's and spheres. For small features, the answer was being dominated by the iterative tolerance and producing wildly problematic values. The logical thing to do is to perform sphere-box collisions using knowledge of the primitives.

This commit adds the following:
  - A new test illustrating the error of GJK is used (see test_fcl_sphere_box.cpp)
  - Borrows the CompareMatrices functionality from Drake and adds it to FCL.
  - Adds the custom sphere-box collision (sphere_box.h and sphere_box-inl.h)
  - Adds *extensive* unit tests on the custom algorithm.
  - Ties the custom algorithm into the libccd and indep GJK solvers.

This PR also includes an independent commit that allows the GJK solvers to be built with `float`.
This PR *doesn't* include the `distance` version of the test in `test_fcl_sphere_box.cpp`. Instead, it contains a TODO.

Sorry for the size, you can chalk that up to the tests (more than half).

```
Category            added  modified  removed  
----------------------------------------------
code                780    1         0        
comments            503    4         0        
blank               157    0         0        
----------------------------------------------
TOTAL               1440   5         0  
```

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/316)


SeanCurtis-TRI:
Merging with single, known CI error (test_fcl_box_box in mac release).

PR #318, opened 2018-07-26
Change transform mode from AffineCompact to Isometry

### LINKED ISSUE #317: Inverting Transform<S> without extra work
The `fcl::Transform` type is an [alias for the `Eigen::Transform` with the `Eigen::AffineCompact`](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/common/types.h#L89). When calling `inverse()` on that method, the `Affine` designation requires that the `linear()` portion have its inverse [computed explicitly](https://bitbucket.org/eigen/eigen/src/0b88199f5b4a7cf10f19855dec234fb321220007/Eigen/src/Geometry/Transform.h?at=3.0&fileviewer=file-view-default#Transform.h-1124).  If the transform generally represents a "rigid" transformation (i.e., rotation and translation), then the inverse of the linear portion should simply be its transpose. This code path is only available if the Transform "mode" is declared to be "Isometry". The problem is we do *a lot* of transforming geometries into other frames by computing inverses.

The `inverse()` method can take a "hint" indicating what *should* be done. I.e., if I know that my transform is strictly a rotation and translation, I can force it to transpose the rotation by passing in `Eigen::Isometry`. It seems it would be good to empower users to do this, based on their certain knowledge.

I'd propose we add a common `fast_inverse` method that essentially does this (and is documented in its assumptions). Something along the lines of:

```c++
/// Creates the inverse of the transform by assuming that the transform is strictly a
/// rotation and translation. The inverse of the rotation is strictly its transpose. This
/// should only be called if the given transform is properly a concatenation of *only*
/// rotation and translation. 
/// @param X_AB   A transform mapping the relationship between frame A and B.
/// @retval X_BA   The inverse of X_AB⁻¹ = X_BA such that X_BA * X_AB = I.
template <typename S>
Transform<S> fast_inverse(const Transform<S>& X_AB) {
   return X_AB.inverse(Eigen::Isometry);
}
```

Issues to consider
  - Users can really screw themselves if they are incorrect in their belief that their X_AB's linear() component is an orthonormal matrix. Computing the inverse generally will give us the most consistent guarantee that `X_BA * X_AB = I`
  - We should also provide some independent test/measure that suggests to them that their transform is "good enough" to be used in a fast transform.

Thoughts?

SeanCurtis-TRI:
This increases the memory footprint of the `Transform` class by 33% (ie.,
3x4 to 4x4). However, the X_AB.inverse() method will now implicitly use
the transpose of the linear() component rather than computing an inverse.

Resolves #317

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/318)


wxmerkt:
Would it make sense to update the [CHANGELOG](https://github.com/flexible-collision-library/fcl/blob/master/CHANGELOG.md)?

jslee02:
> We've got release tags that don't monotonically increase.
> And the changelog implies that all upcoming changes are applied to version 0.6.0. Despite the fact that the 0.5.0 release dated 2016-07-19 has been followed by 0.3.3 on 2016-07-22 and 0.3.4 on 2017-07-21.

They actually monotonically increases but in different branches. A version number is decided based on which branch the change is merged to. For example, changes merging into (current) master would be for 0.6.0, while changes for `fcl-0.5` would be for the next patch release of 0.5 (e.g., 0.5.1 since 0.5.0 is already released).

Sometimes we need to fix a bug for a released version (e.g., 0.3.4) before the next minor or major release (e.g., 0.6.0). Then we might see 0.3.5 release before 0.6.0, which might seem not monotonic increase.

> The changelog seems to have gotten quite stale.

Yeah, this is sad.

> In fact, in adding this change to the changelog, it follows the comment:
> 
>> Switched to Eigen for math oeprations #96, #150

I think that's fine because 0.6.0 hasn't released. So it would be like:
```markdown
Switched to Eigen for math oeprations #96, #150, #318
```
By the way, the typo needs to be fixed! :sweat_smile: 

> It seems we need to make sure the changelog gets updated w.r.t. PRs. It also seems that we need to have a flag day to make sure the changelog is meaningful.

:+1:

[I found a useful site about changelog](https://keepachangelog.com/en/1.0.0/). We don't need to exactly follow the suggestions, but I mostly agree with.

jslee02:
I believe the CI failure is fixed by https://github.com/flexible-collision-library/fcl/pull/266. So it should be gone once `master` is merged.

peci1:
I guess many users will get no improvement at all as the automatic optimization hasn't still been released in Eigen - my comment from https://github.com/ros-planning/geometric_shapes/issues/125#issuecomment-599128522 still holds, it hasn't even got into the recently released Eigen 3.3.8. In Moveit, we took the way of replacing all appropriate `.inverse()` calls to `.linear()`.

SeanCurtis-TRI:
@peci1 The link provided seems to discuss `rotation()` vs `linear()`. In contrast, this is about the cost of `inverse()` (on the full transform). It's true that the choice was made based on the *documented* feature and hasn't been confirmed as you did with `rotation()`. Have you looked at the assembly for what `inverse()` does on an `isometry`?

peci1:
Oh, you're right, I got these two mixed up... The code for inverse() actually does examine the transform traits to improve performance for isometries: https://gitlab.com/libeigen/eigen/-/blob/3.3/Eigen/src/Geometry/Transform.h#L1211 . Sorry for the confusion...

peci1:
And the problem with inefficient `rotation()` should not affect FCL, as

https://github.com/flexible-collision-library/fcl/blob/9dba579158109c0164bfe0e8b4a75c16bfc788f6/include/fcl/narrowphase/collision_object-inl.h#L156-L159

uses the always efficient `linear()` method.

SeanCurtis-TRI:
@peci1 No apologies required. Those kinds of things can fall through the cracks all the time. Worst case scenario, it doesn't *directly* apply, but it's still relevant and interesting (as in this case). More typically, users keeping an eye out for things will catch unintentionally bad code. So, I appreciate you weighing in.

PR #321, opened 2018-07-31
Add primitive cylinder-sphere collision and distance query

SeanCurtis-TRI:
Add custom sphere-cylinder collision and distance tests

By default, the GJK solver was being used for performing distance queries between cylinders and spheres. For small features, the answer was being dominated by the iterative tolerance and producing wildly problematic values. The logical thing to do is to perform sphere-cylinder collisions using knowledge of the primitives.

This commit adds the following:
  - A new test illustrating the error of GJK is used
    (see test_fcl_sphere_cylinder.cpp)
  - Adds the custom sphere-cylinder collision/distance
    (sphere_cylinder.h and sphere_cylinder-inl.h)
  - Adds *extensive* unit tests on the custom algorithm.
  - Ties the custom algorithm into the libccd and indep GJK solvers.

NOTE: Not for review until after #316 merges.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/321)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * Narrowphase
 
   * Add custom sphere-box collision and distance algorithms for both solvers: [#316](https://github.com/flexible-collision-library/fcl/pull/316)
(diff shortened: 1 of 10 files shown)


amcastro-tri:
are we ready to merge?

Issue #322, opened 2018-08-02
Getting assert failure on isOutsidePolytopeFace box to convex hull mesh (sphere)


Levi-Armstrong:
Setup:

- Have a 1m x 1m x 1m box located at <0.0, 0.0, 0.0>
  - defined using fcl::boxd type
- Convex hull of meshed sphere [sphere128.stl.txt](https://github.com/flexible-collision-library/fcl/files/2251842/sphere128.stl.txt) located at <0.2, 0.0, 0.0>
  - defined using fcl::convexd type

When checking these two shapes I am recieving an assert failure in gjk_libccd-inl.h

**Assert**
``` c++
assert(isOutsidePolytopeFace(&polytope, &f, &query_point));
```

**Function**
``` c++
static void ComputeVisiblePatch(
    const ccd_pt_t& polytope, ccd_pt_face_t& f,
    const ccd_vec3_t& query_point,
    std::unordered_set<ccd_pt_edge_t*>* border_edges,
    std::unordered_set<ccd_pt_face_t*>* visible_faces,
    std::unordered_set<ccd_pt_edge_t*>* internal_edges) {
  assert(border_edges);
  assert(visible_faces);
  assert(internal_edges);
  assert(border_edges->empty());
  assert(visible_faces->empty());
  assert(internal_edges->empty());
  assert(isOutsidePolytopeFace(&polytope, &f, &query_point));
  visible_faces->insert(&f);
  for (int edge_index = 0; edge_index < 3; ++edge_index) {
    ComputeVisiblePatchRecursive(polytope, f, edge_index, query_point,
                                 border_edges, visible_faces, internal_edges);
  }
  assert(ComputeVisiblePatchRecursiveSanityCheck(
      polytope, *border_edges, *visible_faces, *internal_edges));
}
```

This was found when testing PR #288. I am still learning my way around fcl so does anyone have any insight into where I should start looking?

Levi-Armstrong:
The previous example was ran against a convex hull sphere with 128 triangles where I just ran it against a convex hull sphere 1028 triangles. In debug it still fails at the assert but it does produce the correct answer. Now thinking back I did some looking into this a while back I found that when two objects are in collision the nearest point returned is the closest vertex which explains why I am getting the correct answer with a higher resolution mesh. @SeanCurtis-TRI and @hongkai-dai is the assert a valid check?

 Calculated Results:

    distance: -0.5499999
    nearest point on box <0.5, -9e-8, 1e-7>
[...]

PR #324, opened 2018-08-03
Add a guard to `faceNormalPointingOutward` to detect degeneracy

SeanCurtis-TRI:
This adds the `triangle_area_is_zero` test to `faceNormalPointingOutward` (with `assert` effect).
This is a prerequisite to the `isOutsidePolytopeFace` test. The test is
rendered meaningless if the reported normal is defined by numerical noise.
If a scenario is failing due to the `isOutsidePolytopFace` assertion
failing, this will detect if it's due to face degeneracies.

relates to #322

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/324)

### DIFF
--- a/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
+++ b/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
@@ -765,6 +765,64 @@ static int simplexToPolytope4(const void *obj1, const void *obj2,
     return 0;
 }
 
+/** Reports true if p and q are coincident. */
+static bool are_coincident(const ccd_vec3_t& p, const ccd_vec3_t& q) {
+  // This uses a scale-dependent basis for determining coincidence. It examines
+  // each axis independently, and only, if all three axes are sufficiently
+  // close (relative to its own scale), are the two points considered
+  // coincident.
+  //
+  // For dimension i, two values are considered the same if:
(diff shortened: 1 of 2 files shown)


Levi-Armstrong:
@SeanCurtis-TRI I have several test in my new environment library and when I check a box to cylinder I receive the following error. My plan is to push the test back to FCL which are checking each shape to another with different conditions.

The box is 1m x 1m x 1m at <0,0,0>.
The cylinder is <0.25,0.25> at <0.2,0,0>

``` bash
tesseract_collision_box_cylinder_unit: /home/larmstrong/catkin_ws/trajopt_public_ws/src/tesseract-1/tesseract_ext/fcl_ros/fcl/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h:848: ccd_vec3_t fcl::detail::libccd_extension::faceNormalPointingOutward(const ccd_pt_t*, const ccd_pt_face_t*): Assertion `!triangle_area_is_zero(a, b, c)' failed.
```

SeanCurtis-TRI:
That's good to hear. And thanks, @Levi-Armstrong for the very specific example. Nice to know we can recreate this problem without having to create a `Convex`. So, the problem (or at least one problem) is that we are generating degenerate faces. 

@hongkai-dai this doesn't seem overly surprising. Something to look into.

PR #325, opened 2018-08-04
Clean up Convex geometry (including fixing AABB computation error)

### LINKED ISSUE #248: When creating Convex Shape the aabb is always inf.

SeanCurtis-TRI:
The Convex geometry had an error in computing its local AABB. This fixes that error.

It includes a unit test that exposes the error without the fix and validates the fix.

This also, incidentally, cleans up some of the documentation.

NOTE: The unit tests on the `Convex` geometry are incomplete. This is noted in the new tests.
replaces #250 (this is part of that PR).

resolves #248 

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/325)

### DIFF
--- a/include/fcl/geometry/shape/convex.h
+++ b/include/fcl/geometry/shape/convex.h
@@ -34,6 +34,7 @@
  */
 
 /** @author Jia Pan */
+/** @author Sean Curtis (2018)  Streamline API and document. */
 
 #ifndef FCL_SHAPE_CONVEX_H
 #define FCL_SHAPE_CONVEX_H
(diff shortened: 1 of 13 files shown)


include/fcl/geometry/shape/convex.h
```diff
@@ -51,7 +51,37 @@ class FCL_EXPORT Convex : public ShapeBase<S_>
...
+  /// This requires *two* coordinated representations of the convex hull:
+  ///   - The set of half spaces that bound the hull region and
+  ///   - the polytope mesh formed by the intersection of the half spaces.
+  ///
+  /// The half spaces are defined by the implicit equation of a plane:
+  /// `Π(x, y, z) = Ax + By + Cz + d = 0` where `n̂ = [A, B, C]` and |n̂| = 1.
+  /// The normal points *outside* of the convex region. There should be *no*
+  /// redundant planes in the set.
+  ///
+  /// For each plane Π there should corresponding face f that lies on that plane
```
Levi-Armstrong: This was one thing I was concerned about. It is possible to have several triangles in the convex mesh that lie on a single plane. So based on this description a user must loop through a mesh and combine all triangle that are on a given plane in to a single polytope. Do I understand this correctly?

include/fcl/geometry/shape/convex.h
```diff
@@ -67,15 +97,35 @@ class FCL_EXPORT Convex : public ShapeBase<S_>
   /// @brief Compute AABB<S>
   void computeLocalAABB() override;
 
-  /// @brief Get node type: a conex polytope 
+  /// @brief Get node type: a convex polytope
   NODE_TYPE getNodeType() const override;
 
-  
+  /// @brief
```
Levi-Armstrong: Missing brief description

Levi-Armstrong:
@SeanCurtis-TRI I built everything using this PR along with #288 and all of my test now pass. :+1:

PR #328, opened 2018-08-27
Toward enabling dashboards on CI

jamiesnape:
Part 1 of many. Relates #329.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/328)


jamiesnape:
I don't think this particular one necessarily needs to wait, however. The others, maybe. I could remove the deprecation warning on the `FCL_BUILD_TESTS` option if you like, but in practice that is not going to be hit by too many people.

Issue #329, opened 2018-08-27
Proposed FCL build/technical debt improvements


sherm1:
Per face-to-face discussion at TRI with @jslee02, @SeanCurtis-TRI, @damrongguoy, and me:

We would like to upgrade FCL development towards modern best-practices. This would include:
- improvements to build system for speed, robustness, and transparency
- set up a build dashboard
- adoption of a uniform coding style guide (likely [Google/Drake](https://drake.mit.edu/styleguide/cppguide.html))
- gradual reduction of technical debt

In the latter category, we are proposing to focus future convex object handling on the built-in GJK/EPA algorithms with the intent of deprecating the libccd dependency eventually. That would allow us to extend Eigen use and differentiability through those algorithms.

We would continue to maintain the current API to the extent possible, with smooth deprecation where necessary.

We would like to hear ASAP from anyone who has further ideas for improvement or concern with this plan.

/cc @jamiesnape

jamiesnape:
It may be a little disruptive, but I would also propose dropping the artificial separation of headers and sources into separate directories `include` and `src`. It makes developing a modular, modern CMake build system rather more complicated.

jamiesnape:
Note that CMake also now has better support for creating Debian and Ubuntu compliant packages, so we can automate the release process (relates #331).

sherm1:
> It may be a little disruptive, but I would also propose dropping the artificial separation of headers and sources into separate directories include and src.

I imagine that would be quite disruptive for FCL users (any users care to weigh in?). I'm guessing that many work with the source directory structure rather than an installed FCL (where headers could be relocated). How important is it to reorganize these, @jamiesnape -- I would think CMake could be made to deal with it as is.

jamiesnape:
Installed or otherwise is completely separate to this, so not quite sure what you mean. I would have thought having the sources next to the relevant headers would be distinctly easier for developers.

jamiesnape:
Note also that headers are not really relocated modulo the name of the top-level directory may change. If you pointed at `fcl/include` before, pointing to `fcl` would have the same effect. It is the source files that move.

jamiesnape:
>  I would think CMake could be made to deal with it as is.

It can, but that does not mean it should. You will have more maintainable code and a better user experience with a one-off merge of directories, and it does not even change the API.

SeanCurtis-TRI:
[...]

PR #332, opened 2018-09-04
Add configuration files for various static analyzers

jamiesnape:
Part 2 of CI-related work. This assumes you choose some derivative of the Google C++ Style Guide in #329.


SeanCurtis-TRI:
We might want to stall slightly on this. We've got an outstanding bug (apparently) with the EPA implementation. It would make sense to patch that, tag and release a new version of fcl before we start playing with the infrastructure (i.e., #329 related stuff).

@hongkai-dai how credible do we feel the EPA bug is? Last we discussed it, it felt like we don't have a way to reproduce it. If that's the case, we can move on with a new release. Otherwise, we should probably move to release FCL so we can start tackling #329.

jamiesnape:
I assuming most of these are going to wait for a while. No rush.

sherm1:
@SeanCurtis-TRI can we move ahead with this yet?
@jamiesnape (needs a manual merge/rebase)

jamiesnape:
Actually running the tools and fixing the defects is, I imagine, going to be after the next release, but I would assert that this is fine to review and potentially merge.

I would also say this is really a starting point for the configurations to have something with which to develop CI based on my best guess for the project's requirements. If the configurations need to change now or prior to them being enforced, it is fine by me.

Issue #326, opened 2018-08-07
Convex geometry clean up


SeanCurtis-TRI:
Perusing the [`Convex`](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/geometry/shape/convex.h) geometry implementation, several issues jump out:

1. The class has several unused members. They aren't used in the class's internal implementation and, it likewise appears, are not directly referenced (as public members). They should be removed from the class API.
    - `plane_normals`
    - `plane_dis`
    - `edges`
2. The class does no validation on the inputs. It can either validate by default at construction or be available as a user-callable a method such that the user can choose to validate.
3. The operations that calculate volume, center of mass, and moment of inertia tensor all have duplicated geometry (they all rely on the volume-weighted combination of smaller tetrahedra). To iterate through the tetrahedra and compute volume is *duplicated* across all methods. Refactor so it's only defined once.
4. Currently, the smaller tetrahedra are defined generally (every face is divided into tetrahedra based on the face's "center" point and a fan of triangles including all other vertices). If the face is actually a triangle, this leads to volume for three tetrahedra being computed instead of just the one. The decomposition of the geometry into tetrahedra should, at least, handle triangular faces more intelligently. And perhaps also 4-sided faces.
5. There are no (until #325 and then limited) unit tests for the geometry.

SeanCurtis-TRI:
@jslee02 Do you know any reason why I can't pull the trigger on modifying the API of the `Convex` type?

jslee02:
@SeanCurtis-TRI Nope. I think your points are all valid. Looking at the history, `Convex` class was added in the very beginning of FCL (7 years ago), and there hasn't been any significant improvement made. It's time to move on!

As a note, one of my personal reasons I haven't taken a look at this class is that I believed there is no performance advantage of using `Convex` over the BVH representation even for convex meshes. This is because FCL uses the convexity-based algorithms for `Convex`, and [the specific algorithm frequently requires iteration over all the vertices](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L2106). This conclusion is made without a comprehensive benchmark so it could be wrong. However, the current algorithm can be replaced by a more efficient algorithm anytime so I think this modification worth enough.

SeanCurtis-TRI:
Hi @jslee02 Thanks for weighing in.

[...]

PR #334, opened 2018-09-07
Add convexhull to convehull test 

Levi-Armstrong:
This test triggers the assert for isOutsidePolytopeFace in ComputeVisiblePatch

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/334)

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -63,6 +63,7 @@ set(tests
     test_fcl_sphere_capsule.cpp
     test_fcl_sphere_cylinder.cpp
     test_fcl_sphere_sphere.cpp
+    test_fcl_convexhull_convexhull.cpp
 )
 
 if (FCL_HAVE_OCTOMAP)
--- a/test/test_fcl_convexhull_convexhull.cpp
+++ b/test/test_fcl_convexhull_convexhull.cpp
@@ -1,0 +1,412 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2014-2016, CNRS-LAAS and AIST
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of CNRS-LAAS and AIST nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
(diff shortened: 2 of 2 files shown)


Levi-Armstrong:
This is to assist with issue #326 testing

Levi-Armstrong:
Here is th ply files if you want to view the convex hull for commit ab76114
[link_7_0.ply.txt](https://github.com/flexible-collision-library/fcl/files/2359493/link_7_0.ply.txt)
[sphere_link_0.ply.txt](https://github.com/flexible-collision-library/fcl/files/2359496/sphere_link_0.ply.txt)

hongkai-dai:
Hi Levi,

Thanks a lot for the PR! I think we can fix the problem in that assert with LIBCCD solver. I will file a PR with part of your test to expose the problem.

Levi-Armstrong:
I have added second of three tests which triggers another assert.

hongkai-dai:
Hi Levi,

I believe #336 should fix the problem `convexhull_convexhull_ccd`.

`convexhull_convexhull_indep` calls a totally different code path. It uses FCL's independent solver with `collide` function. It incorrectly returns a negative value as the penetration depth. I do not think I can quickly fix that code  (I didn't touch `collide` function or `indep` solver before). Would you mind to comment out `convexhull_convexhull_indep` test, and add a `TODO` in the documentation, saying that we will fix this test later?

I will look into the `box_cylinder_test`. It triggers `!triangle_area_is_zero(a, b, c)` assert in `faceNormalPointingOutward`. This is likely a degenerate case. @SeanCurtis-TRI 

Best,
Hongkai

Levi-Armstrong:
@hongkai-dai I commented out the one test. I also added another test for different scenarios related to checking two convex hulls. Looking at the results of the test it has some large errors when two objects are in collision for distance and nearest points. The following case are currently tested for when the closest feature is:

- Edge to Edge
- Vertex to Vertex

Need to add:
- Edge to Vertex
- Edge to Face
- Vertex to Face
- Face to Face

 to Edge

hongkai-dai:
Hi Levi,

Sorry for my belated reply.

I am looking into your test in https://github.com/Levi-Armstrong/fcl/blob/convexTests/test/test_fcl_convexhull_convexhull.cpp#L542L546. May I ask you you obtain these numbers `-0.27552, 0.23776, -0.03776`?

In your comment https://github.com/Levi-Armstrong/fcl/blob/convexTests/test/test_fcl_convexhull_convexhull.cpp#L529, you mentioned that the closest feature between the two colliding objects are edge to edge. I think mathematically it is proved that the closest point on the boundary of the Minkowski difference A ⊖ B is an interior point within each facet (see theorem 4.8 of https://github.com/tianxiao/Recent-Advances-in-Real-Time-Collision-and-Proximity-Computations-for-Games-and-Simulations-/blob/master/books/Collision%20Detection%20in%20Interactive%203D%20Environments.pdf). Namely if your polytope is well posed (no two neighbouring faces are parallel), then the closest feature can never be edge to edge, at least one of the feature should be a face.

hongkai-dai:
Hi Levi,

Sean @SeanCurtis-TRI and I looked into your test on the closest distance. We think in https://github.com/Levi-Armstrong/fcl/blob/convexTests/test/test_fcl_convexhull_convexhull.cpp#L542L546, what you find is not the penetration depth, but the distance between two edges.

The definition of the penetration depth is 
min |v|
s.t (A + v) ∩ B = ∅

Namely the penetration depth is the minimal translational distance, such that the two objects A and B can get out of contact. While the distance you find can get the two objects out of contact, it is not the minimal one.

To show the minimal translation distance, Sean draws the two polytopes as
spheres_dense_colliding

We draw the two closest points `p1`, `p2` found from EPA
sphere_transparent_colliding

And we translate one sphere with a vector `p1 - p2`, here are the translated spheres. You can see that they are separated
spheres_dense_separating

The distance found from EPA is 0.2706222, smaller than the distance you computed (0.27552). So by definition of penetration depth, the EPA distance is more correct than 0.27552.

Note that the closest points found from EPA are not on the edges, and the vector `p1 - p2` is not along the x axis.

Levi-Armstrong:
@hongkai-dai and @SeanCurtis-TRI  This is great and thank you for the detailed response. I came up with the number using blender and physically measuring two meshes. The one difference is I was measuring along the the same translation vector provided by using primitive shapes assume they should be same. As you pointed out this will not be the same so bad assumption on my part. I will keep looking.

Levi-Armstrong:
So I pulled the meshes into blender and transformed one mesh 0.2m in the x direction. Then created a line between the nearest points(Provided by FCL) to understand where it was on the mesh. Based on your description it sounds like the closest objects should not be two edges but when visualizing it looks like they are both on an edge. Although they could be just very close to the edge but it just looked suspicious. I am going to calculate how close it is and post the results.

Point: -0.025420749082411945, -0.06229177389074456, -0.025420749082411945
image

Point: -0.032902129264164116, -0.06229177389074456, 0.025420749082411945
image

PR #336, opened 2018-09-10
Fix face normal pointing outward

hongkai-dai:
Fix two problems in `faceNormalPointingOutward`

1. One problem is exposed in @Levi-Armstrong's PR #334, that when we find a vertex as a good witness point that can determine the direction of the face normal, we should return directly, instead of break.
2. When computing the projection from a vertex to the plane, we incorrectly computed the projection to the origin.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/336)


SeanCurtis-TRI:
Looks like mac and windows are a bit less permissive in your use of `std::array`. Try doing a `#include <array>` -- that should make the CI errors go away.

I'll look at the PR in more detail in the morning.

Levi-Armstrong:
I ran this version for multiple motion plans and I have not had an assert failure for faceNormalPointingOutward.

hongkai-dai:
I don't understand the appveryor build failure, have we seen this kind of failure before?

PR #338, opened 2018-09-17
Change convex shape constructor

Levi-Armstrong:
Based on comments related to the convex shape to take `std::shared_ptr<aligned_vector<Vector3<S>>>` and `std::shared_ptr<std::vector<int>>` I decided to move forward and make the changes. 

I originally went down the route of changing all `Vector<S>*` to `const aligned_vector<Vector3<S>>&` but this quickly took me down a rabbit hole. I have a different branch where I left off, so if this is desirable I will continue working it and create a pull request. This PR does the minimum to integrate the use of new types.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/338)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,10 @@
 
 * Geometry
 
-  * Simplified Convex class, deprecating old constructor in favor of simpler, documented constructor: [#325](https://github.com/flexible-collision-library/fcl/pull/325)
+  * Simplified Convex class, deprecating old constructor in favor of
+    simpler, documented constructor:
+     [#325](https://github.com/flexible-collision-library/fcl/pull/325),
+     [#338](https://github.com/flexible-collision-library/fcl/pull/338)
 
 * Broadphase
 
--- a/include/fcl/geometry/shape/convex-inl.h
+++ b/include/fcl/geometry/shape/convex-inl.h
@@ -61,7 +61,9 @@ Convex<S>::Convex(const std::shared_ptr<const std::vector<Vector3<S>>>& vertices
   // Compute an interior point. We're computing the mean point and *not* some
   // alternative such as the centroid or bounding box center.
   Vector3<S> sum = Vector3<S>::Zero();
-  sum = std::accumulate(vertices_->begin(), vertices_->end(), sum);
+  for(const auto& vertex : *vertices_) {
+    sum += vertex;
+  }
   interior_point_ = sum * (S)(1.0 / vertices_->size());
 }
 
--- a/include/fcl/geometry/shape/convex-inl.h
+++ b/include/fcl/geometry/shape/convex-inl.h
@@ -120,18 +120,10 @@ Matrix3<S> Convex<S>::computeMomentofInertia() const {
     // Compute the volume of tetrahedron formed by the vertices on one of the
     // polygon's edges, the center point, and the shape's frame's origin.
     const Vector3<S>& v3 = face_center;
-    for (int j = 1; j <= vertex_count; ++j) {
-      // Each edge of the polygon is defined where faces[face_index + j] is
-      // the first edge vertex and faces[face_index + (j % vertex_count) + 1]
-      // is the second edge vertex. The use of (j % vertex_count) is required
-      // to handle the last polygon edge which is defined by the last polygon
-      // vertex and the first vertex. When j < vertex_count then
-      // (j % vertex_count) == j resulting in faces[face_index + j + 1] which
-      // is the next polygon vertex. When j == vertex_count then
-      // (j % vertex_count) == 0 resulting in faces[face_index + 1] which is
-      // the first vertex of the polygon.
-      int e_first = faces[face_index + j];
-      int e_second = faces[face_index + (j % vertex_count) + 1];
+    const int vertex_base = face_index + 1;
+    for (int j = 0; j < vertex_count; ++j) {
+      int e_first = faces[vertex_base + j];
+      int e_second = faces[vertex_base + (j + 1) % vertex_count];
       const Vector3<S>& v1 = vertices[e_first];
       const Vector3<S>& v2 = vertices[e_second];
       S d_six_vol = (v1.cross(v2)).dot(v3);
--- a/include/fcl/geometry/shape/convex-inl.h
+++ b/include/fcl/geometry/shape/convex-inl.h
@@ -50,22 +50,19 @@ class FCL_EXPORT Convex<double>;
 
 //==============================================================================
 template <typename S>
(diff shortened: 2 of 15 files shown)


Levi-Armstrong:
@SeanCurtis-TRI Should the member variables be made private and then provide methods for retrieving the data?

SeanCurtis-TRI:
In the absence of any data suggesting otherwise, I'd say yes. As I was going over the PR, I contemplated raising that issue. But since I'd quibbled about so much, and the public members was part of the FCL API already, I was willing to look the other way. But proper encapsulation would definitely be a good thing.

Levi-Armstrong:
Good. I am about to push another commit addressing all of the requested changes except the validate function which I am working on now. I remember now why I did not make it const objects because it required several function to be updated to allow the use of const objects. I have made the changes to allow the use of const objects for the convex shape.

Also I was not sure on how to best name the method for getting the number of faces. Let me know if you would like to name it something else.

Levi-Armstrong:
___
*[test/geometry/shape/test_convex.cpp, line 98 at r2](https://reviewable.io/reviews/flexible-collision-library/fcl/338#-LMh2Loh4UJ0ssviBNkR:-LMhIvz84dEOmxkRX5jj:b-5312ik) ([raw file](https://github.com/flexible-collision-library/fcl/blob/c9d9143d4c0f131f8885caf776c21f85b6086862/test/geometry/shape/test_convex.cpp#L98)):*

After a quick experiment, it looks like this would work:

```c++
return std::unique_ptr<const std::vector<Vector3<S>>>>(vertices_);
```

I did a quick test and it appears that it properly shares the underlying structure.
</blockquote></details>

Thank you for pointing me to the right file. I should not have been returning const reference to the shared pointer for the Polytope Class.

Levi-Armstrong:
Ok, so everything builds now but tests are failing so I will look into this further.

SeanCurtis-TRI:
That is an *excellent* point. However, I saw we proceed for the following reasons:

1. This doesn't change the last releases. So, people who are using a fixed release will be unaffected.
2. We've already touched the Convex interface (at least construction) recently in master and have heard no complaints. So, this follow up change is *probably* safe.

Levi-Armstrong:
@SeanCurtis-TRI I found a few issues that I introduce which have been resolved but I can't figure out what I have changed that would causing a few of the tests to fail. Do you have any insight?

Levi-Armstrong:
@SeanCurtis-TRI I did some more investigating and I am still not sure why the two test are failing using float types.

Levi-Armstrong:
I decided to try the original function code where I convert the `std::vector` to c_array to see if maybe I made an error translating the function but I get the same test failures. Not sure what to do now. Overall the error is still very small and it is scaling the test tolerance based on the volume size which does make sense but is it an accurate representation of the expected error?

```  S vol = 0;
  std::vector<Vector3<S>> temp_vertices(*vertices_);
  std::vector<int> temp_faces(*faces_);
  Vector3<S>* vertices = &temp_vertices[0];
  int *face_encoding = &temp_faces[0];
  int *index = face_encoding + 1;
  for(int i = 0; i < num_faces_; ++i) {
    const int vertex_count = *face_encoding;
    Vector3<S> face_center = Vector3<S>::Zero();

    // TODO(SeanCurtis-TRI): While this is general, this is inefficient. If the
    // face happens to be a triangle, this does 3X the requisite work.
    // If the face is a 4-gon, then this does 2X the requisite work.
    // As N increases in the N-gon this approach's inherent relative penalty
    // shrinks. Ideally, this should at least key on 3-gon and 4-gon before
    // falling through to this.

    // Compute the center of the polygon.
    for(int j = 0; j < vertex_count; ++j)
      face_center += vertices[index[j]];
    face_center = face_center * (1.0 / vertex_count);

    // TODO(SeanCurtis-TRI): Because volume serves as the weights for
    // center-of-mass an inertia computations, it should be refactored into its
    // own function that can be invoked by providing three vertices (the fourth
    // being the origin).

    // Compute the volume of tetrahedron formed by the vertices on one of the
    // polygon's edges, the center point, and the shape's frame's origin.
    const Vector3<S>& v3 = face_center;
    for(int j = 0; j < vertex_count; ++j) {
      int e_first = index[j];
      int e_second = index[(j + 1) % vertex_count];
      const Vector3<S>& v1 = vertices[e_first];
      const Vector3<S>& v2 = vertices[e_second];
      S d_six_vol = (v1.cross(v2)).dot(v3);
      vol += d_six_vol;
    }

    face_encoding += vertex_count + 1;
    index = face_encoding + 1;
  }

  return vol / 6;
}
```

SeanCurtis-TRI:
I didn't get a chance to look at it today. I'll look at it first thing in the morning to see if I can come up with anything.

Levi-Armstrong:
That make since. I am curious, how did you track this down?

The test is now passing and thanks for the help tracking this down. It looks like CI has an issue with using std::accumulate on clang XCode.

SeanCurtis-TRI:
Once I'd convinced myself the `std::accumulate` wasn't making a difference, the only thing I could suppose is that *somehow* the order of operations was different. So, I applied logging to the volume computation that would spew out indices, vertex values, volume values, etc, during its computation. After I did that, I wanted to reduce the spew and only do the *single* test that was failing (with the cube, that is). And I found that the test in question had *different* vertex values when I ran it in sequence vs running it alone. That's when I realized that the values were being changed which immediately led to the shared-ptr, deep/shallow copy shenanigans. 

If we have to, to appease the Mac gods, we can just go back to a for loop. Not the end of the world.

Levi-Armstrong:
@SeanCurtis-TRI I removed the use of std::accumulate and CI is happy now.

Levi-Armstrong:
*[a discussion](https://reviewable.io/reviews/flexible-collision-library/fcl/338#-LMbtg_r-OeToLP0dw7p:-LNC24oY0aS4j6b9I6Lw:b8xjx8y) (no related file):*

On a completely different subject -- I presume you're using github to do the review, yes? There's a "Reviewable" link in the PR and I highly recommend trying it out. It's a much cleaner interface for having conversations and tracking issues. Let me know if you want a run down on it.
</blockquote></details>

I have been going through the process but would be interested to know how best to use it. I went through and selected the acknowledged, Resolved, or Done. 

---

Levi-Armstrong:
Sorry I have been out of vacation. I will work on this today.

Levi-Armstrong:
I plan to make the remaining changes this week and I apologize for the delay.

Levi-Armstrong:
Allow edits from maintainer is enabled.

Issue #361, opened 2019-01-17
test_fcl_distance NearestPointFromDegenerateSimplex fails on macOS


jamiesnape:
According to CI. Probably due to a change in the version of a Homebrew dependency than a code change.

```
Test #13: test_fcl_distance ...............................***Failed    3.26 sec
[==========] Running 2 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 2 tests from FCL_DISTANCE
[ RUN      ] FCL_DISTANCE.mesh_distance
distance timing: 0.054579 sec
collision timing: 0.036614 sec
[       OK ] FCL_DISTANCE.mesh_distance (3253 ms)
[ RUN      ] FCL_DISTANCE.NearestPointFromDegenerateSimplex
/Users/user/fcl/test/test_fcl_distance.cpp:363: Failure
The difference between expected_dist and result.min_distance is 2.3454265710154409e-07, which exceeds constants<ccd_real_t>::eps_34(), where
expected_dist evaluates to 0.053516322172152138,
result.min_distance evaluates to 0.053516087629495036, and
constants<ccd_real_t>::eps_34() evaluates to 1.8189894035458565e-12.
[  FAILED  ] FCL_DISTANCE.NearestPointFromDegenerateSimplex (0 ms)
[----------] 2 tests from FCL_DISTANCE (3253 ms total)

[----------] Global test environment tear-down
[==========] 2 tests from 1 test case ran. (3253 ms total)
[  PASSED  ] 1 test.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] FCL_DISTANCE.NearestPointFromDegenerateSimplex

 1 FAILED TEST
```

jamiesnape:
(Will post more details when I have a chance to do a local build.)

SeanCurtis-TRI:
Further thoughts/notes/observations:

Starting [Jan 16, 2019](https://travis-ci.org/flexible-collision-library/fcl/builds/480449098), we've introduced a new, persistent failure in the mac CI builds. Specificaly:

test_fcl_distance.cpp, NearestPointFromDegenerateSimplex(). This test arose from a bug reported in #293 and fixed by #296. However, the test is documented as having some very specific requirements which may or may not be met. 

[...]

PR #360, opened 2019-01-16
Various CMake fixes and cleanup

jamiesnape:
Do not set redundant macOS-related policy; `CMP0042` is enabled by virtue of requiring CMake 3.0 on macOS. Do not require a C compiler; there is no C code. Fix typo in `CMAKE_INSTALL_FULL_LIBDIR` variable name; there is no such variable `CMAKE_INSTALL_LIBDIR_FULL`. Remove redundant set of `CMAKE_SHARED_MODULE_PREFIX`; there are no modules being built.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/360)

PR #352, opened 2018-11-17
Fix the scalar type in a test.

hongkai-dai:
Resolves #349

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/352)

PR #365, opened 2019-01-29
use the actual distance instead of the squared distance to determine if the origin is on the facet of the simplex.

hongkai-dai:
cc @avalenzu

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/365)

### DIFF
--- a/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
+++ b/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
@@ -372,7 +372,7 @@ static int doSimplex4(ccd_simplex_t *simplex, ccd_vec3_t *dir)
   // if it is not simplex can't be expanded and thus no intersection is
   // found
   dist_squared = ccdVec3PointTriDist2(&A->v, &B->v, &C->v, &D->v, nullptr);
-  if (ccdIsZero(dist_squared)) {
+  if (isAbsValueLessThanEpsSquared(dist_squared)) {
     return -1;
   }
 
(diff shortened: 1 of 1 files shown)


sherm1:
I'm confirming that the CI failure is only the test_fcl_distance tolerance problem that we've been seeing on Macs, unrelated to this PR.

SeanCurtis-TRI:
The mac test hasn't failed in ages? Why did it suddenly start failing now?

hongkai-dai:
The last PR https://github.com/flexible-collision-library/fcl/pull/362 merged before this one also has the same mac test failure.

SeanCurtis-TRI:
So, in November (#352) was the last PR that was about math. That passed (and several prior to that). Then we had a couple of CMake-related PRs that failed. Weird that that's come back.

sherm1:
From looking at the closed-PR list it appears that #360 was the first one that failed. That did include some small Mac-only CMake changes. Hard to see how that would break anything though -- any thoughts @jamiesnape ?

jamiesnape:
Not related. Master actually failed before then.

SeanCurtis-TRI:
Precisely - it's the same mac error we've been getting since  January 16 (https://travis-ci.org/flexible-collision-library/fcl/builds/480449098). That is *not* the failure we were consistently getting all of last year where we'd confirm it was the "known failure" and merge anyways.

This is failure in a very specific test -- one related to #293. It requires a box-box distance test to be evaluated with GJK. Curiously, the expected answer and computed answer differ by 10^-7 which feels very theshold-ish; but I'm not sure what would cause a change in behavior. I'll post an issue.

jamiesnape:
> I'll post an issue.

#361 is not that issue?

SeanCurtis-TRI:
@jamiesnape you are 100% right. I've assigned myself to that issue. Thanks.

PR #373, opened 2019-02-12
Literal defined as ccd_real_t

### LINKED ISSUE #372: compilation error in test file
~~~
fcl/test/narrowphase/detail/convexity_based_algorithm/test_gjk_libccd-inl_gjk_doSimplex2.cpp:181:67: note:   deduced conflicting types for parameter ‘_Tp’ (‘double’ and ‘float’)
     const ccd_real_t eps = std::max({1., p_OA.norm(), p_OB.norm()}) * kEps;
~~~

Changing the `1.` to `1.0f` does fix the error, but I'm guessing that the double/float error might be a configuration thing (maybe ?).

SeanCurtis-TRI:
A literal in a type-deduced function call was declared as a `double` literal (e.g., `1.`). When `ccd_real_t` is a `double`, this is fine. But when it is a `float`, the type cannot be successfully deduced.

We correct this by explicitly calling it out as a `ccd_real_t`-typed value.

resolves #372

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/373)


doudou:
Compiles fine here... Thanks

Issue #366, opened 2019-01-30
__ccdGJK in narrowphase has bugs


hongkai-dai:
## Problem
When computing the signed distance in the narrow phase convexity based algorithm, FCL first calls `__ccdGJK` [function](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L1350-L1403) to determine if the object A and B are separated. This code has several bugs

1. There are 3 possible outcomes from this function. 
   - A separating plane between A and B are founded. Hence A and B are not intersecting.
   - The origin is in the Minkowski difference A ⊖ B. Hence A and B are intersecting.
   - We reached the iteration limit without determining whether A and B are intersecting or not.
   On the other hand, there are only two return values of this function.
   - When A and B intersects, return 0
   - Otherwise, return -1.
   Hence the FCL code doesn't differentiate between separating and iterations limit. We should improve the return values to make that difference.
2. `__ccdGJK` lies when it returns "no intersection" in the following lines https://github.com/flexible-collision-library/fcl/blob/9082fd2be665b2245716fdcb22ff5ae6ad29d892/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L1396-L1398 Here when `|dir| = 0`, since `dir = -v` (where `v` is the point that is closest to the origin in the simplex), this means that `v = 0`, and hence the origin is within the simplex, hence the origin is within A ⊖ B. This means A and B are intersecting. FCL returns the wrong code.
3. When calling `doSimplex2` to compute the closest point within the new simplex (line segment `AB`) to the origin `O`, FCL also checks if the origin is on the line segment `AB`. The check it does is https://github.com/flexible-collision-library/fcl/blob/9082fd2be665b2245716fdcb22ff5ae6ad29d892/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L245-L252 Namely if `AB x AO = 0` and `AB.dot(AO) > 0`, then the code thinks that `O` is on the line segment `AB`. I think this check is wrong. It could be that O is on the ray shooting from A to B, but is beyond B. Hence this `O` satisfies the check in FCL, but it is not on the line segment `AB`.
[...]

PR #388, opened 2019-04-06
When libccd reports the nearest feature is an edge, validate the result

### LINKED ISSUE #319: libccd EPA expandPolytope: "this should not happen"
As noted in https://github.com/flexible-collision-library/fcl/pull/305#discussion_r205945024, I have encountered the "should not happen" logic error:
```
RuntimeError: FCL expandPolytope(): This should not happen. Both the nearest point and the new vertex are on an edge, thus the nearest distance should be 0. This is touching contact, and we should not expand the polytope for touching contact.
```

Here are details on how to reproduce:

- ``fcl::distance`` call between a box (o1, GEOM_BOX) and a cylinder (o2, GEOM_CYLINDER).
- The box dimensions are: ``.05 0.39 1.47``
- The cylinder dimensions are: ``length=0.18, radius=0.18``
- The box transform is: ``xyz=0.8 -0.385  0.735; wxyz=0.707107 0 0 0.707107``
- The cylinder transform is:  ``xyz=0.4 -0.4 0.93; wxyz=0.991445 0 0 -0.130526``

I have been able to reliably reproduce it (as part of a larger process - caught and extracted the above from it).

cc @hongkai-dai

hongkai-dai:
Also fixes #319.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/388)

### DIFF
--- a/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
+++ b/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
@@ -41,6 +41,7 @@
 #include "fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd.h"
 #include "fcl/narrowphase/detail/failed_at_this_configuration.h"
 
+#include <array>
 #include <unordered_map>
 #include <unordered_set>
 
--- a/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
+++ b/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h
@@ -395,9 +395,9 @@ static int doSimplex3(ccd_simplex_t *simplex, ccd_vec3_t *dir)
   // Compute origin_projection as well. Without computing the origin projection,
   // libccd could give inaccurate result. See
   // https://github.com/danfis/libccd/issues/55.
-  ccd_vec3_t origin_projection;
+  ccd_vec3_t origin_projection_unused;
   dist = ccdVec3PointTriDist2(ccd_vec3_origin, &A->v, &B->v, &C->v,
-                              &origin_projection);
+                              &origin_projection_unused);
   if (isAbsValueLessThanEpsSquared(dist)){
     return 1;
   }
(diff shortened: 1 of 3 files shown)


wxmerkt:
This fixes #319 - thank you @hongkai-dai :)

wxmerkt:
I am running this branch in an extensive computation process - so far it really improved the stability of the process and did solve #319.

I just encountered the following though:
```
 Original error message: /home/wxm/dev/install_ws/build/fcl_catkin/fcl_src-prefix/src/fcl_src/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h:(1593): validateNearestFeatureOfPolytopeBeingEdge(): The origin is outside of the polytope. This should already have been identified as separating.
  Shape 1: Box0.02999999999999999889 0.11999999999999999556 0.10000000000000000555
  X_FS1
-3.0627937852578681533e-08    -0.99999999999999888978 -2.8893865161583314238e-08     0.63499979627350811029
     0.9999999999999980016 -3.0627939739957803544e-08  6.4729926918527511769e-08    -0.48500002215636439651
-6.4729927722963847085e-08 -2.8893863029448751323e-08     0.99999999999999711342      1.0778146458339641356
                         0                          0                          0                          1
  Shape 2: Box0.025000000000000001388   0.3499999999999999778   1.8449999999999999734
  X_FS2
6.1232339957367660359e-17                        -1                         0    0.80000000000000004441
                        1 6.1232339957367660359e-17                         0   -0.45750000000000001776
                        0                         0                         1     1.0224999999999999645
                        0                         0                         0                         1
  Solver: GjkSolver_libccd
    collision_tolerance:      2.0097183471152322134e-14
    max collision iterations: 500
    distance tolerance:       9.9999999999999995475e-07
    max distance iterations:  1000
```

As this statement has been newly introduced in this PR I am posting it here - but it may be better to place it in a new issue?

Do you have any insights on this?

Thanks,
Wolfgang

hongkai-dai:
Thanks @wxmerkt ! Is it OK if you could post the error in a separate issue? My guess is that the problem occurs in the `__ccdGJK` code (the code that determines whether the two objects are colliding or separating). It probably incorrectly decides that the two geometries are colliding, but they are not. Hence the error.

We have an issue on #366 , @SeanCurtis-TRI submitted a PR #367 to fix some of the bugs that cause the issue #366, but we know that we haven't totally fixed the issue yet.

SeanCurtis-TRI:
My guess is that there are several stumbling blocks. This PR helped you over the earlier stumbling block so you could hit the later stumbling block. Fortunately, we should be able to reproduce the bug given the error report.

sherm1:
FYI I checked the rotation matrix in X_FS1 -- it is perfectly orthonormal so the 1e-08's in it actually represent that Shape1 was reoriented _almost_ exactly 90 degrees but not quite!

hongkai-dai:
I confirm that in @wxmerkt example, `__ccdGJK` gives the wrong polytope to EPA. `__ccdGJK` generates a tetrahedron, and it thinks that the origin is in this tetrahedron (indicating the two boxes are colliding). But here are the four vertices of this tetrahedron
[-0.4, -2.07575e-08, 1.02781]
[0.0699998, -1.70822e-08, -0.817185]
[0.0699998, -1.70822e-08, 1.02781]
[-0.28, -1.70822e-08, 0.105315]
As you can see, all four vertices has negative value along the `y` axis. Hence the whole tetrahedron is on the -y side. This tetrahedron cannot contain the origin.

I will look into the `doSimplex` code to find out the cause of the bug.

PR #413, opened 2019-07-22
Use Eigen3 vectorized operations in obbDisjoint

yszheda:
Hi,

This is Shuai from dorabot. We heavily use OBB BVs in our collision detection procedure and we've found that the OBB intersection checking is one of the performance bottlenecks. So we have tried different ways to optimize this part. One of my observation is that the current implementation of `fcl::obbDisjoint` uses scalar operations for the 15 serial separating-axis tests, which can be vectorized and run faster in CPUs with SIMD support.
This PR is one of our vectorized approaches. We group the 15 separating-axis tests into 5 groups of tests, and use vectorization to better exploit the potential of `Eigen3` acceleration. From our experiments (as I will explain later), we can achieve approximately 10% performance improvement for the `fcl::obbDisjoint<float>`.

## Experiments of Performance Comparison

I have collected 133464 `fcl::obbDisjoint` testcases from our application real cases. I also notice that there might be cases where the original scalar implementation is better than the vectorized one: for example, if the first tested axis is already the separating-axis, the original scalar implementation would immediately `return` while the vectorized implementation would still have to complete checking three tested axes. In order to better see the impact, I group the testcases by the number of tested axes. For each testcase, I also run the test for more than thousands of times to measure the stable latency.

Tested environment:
* CPU: Intel(R) Core(TM) i7-8700K @ 3.70GHz x12
* OS: Linux debian 4.9.0-8-amd64

Here I show you the performance comparison for both `float` and `double` specialization of `fcl::obbDisjoint<>` (The SSE and AVX ones are from another patch, which use SSE and AVX intrinsics for vectorization. I will create another PR for that XD):

fcl_obbDisjoint_float

fcl_obbDisjoint_double

From the results above, we can find that:
* The latency of the original scalar implementation increases faster as the number of tested axes increase.
* If the number of tested axes is small (like 1 or 2), the original scalar implementation would still be better due to the "early-return" behavior. However, its advantage in this best case is inobvious compared to its disadvantage when the number of tested axes increase.
* The Eigen vectorization patch fails to achieve better performance for `fcl::obbDisjoint<double>`. And that's why I decide to use it only in the `float` specialization.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/413)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,12 +42,12 @@ script:
   - cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DFCL_TREAT_WARNINGS_AS_ERRORS=ON -DFCL_COVERALLS=$COVERALLS ..
 
   # Build
-  - make -j4
+  - travis_wait 30 make -j $(nproc)
   - if [ $COVERALLS = ON ] && [ "$TRAVIS_OS_NAME" = "linux" ]; then make coveralls; fi
 
   # Run unit tests
-  - ctest -j4 --output-on-failure
+  - travis_wait 30 ctest -j $(nproc) --output-on-failure
 
   # Make sure we can install and uninstall with no issues
-  - sudo make -j4 install
-  - sudo make -j4 uninstall
+  - sudo make -j $(nproc) install
+  - sudo make -j $(nproc) uninstall
--- a/include/fcl/math/bv/OBB-inl.h
+++ b/include/fcl/math/bv/OBB-inl.h
@@ -100,9 +100,7 @@ bool OBB<S>::overlap(const OBB<S>& other) const
   /// compute the relative transform that takes us from this->frame to
   /// other.frame
 
-  Vector3<S> t = other.To - To;
-  Vector3<S> T(
-        axis.col(0).dot(t), axis.col(1).dot(t), axis.col(2).dot(t));
+  Vector3<S> T = axis.transpose() * (other.To - To);
   Matrix3<S> R = axis.transpose() * other.axis;
 
   return !obbDisjoint(R, T, extent, other.extent);
@@ -522,6 +520,10 @@ bool obbDisjoint(const Matrix3<S>& B, const Vector3<S>& T,
 
 }
 
+template <>
+bool obbDisjoint(const Matrix3<float>& B, const Vector3<float>& T,
+                 const Vector3<float>& a, const Vector3<float>& b);
+
 //==============================================================================
 template <typename S>
 bool obbDisjoint(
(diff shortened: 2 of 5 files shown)


yszheda:
Hi @sherm1,

Thanks for the reviewing and suggestion!

Personally I would favor the patch on https://github.com/flexible-collision-library/fcl/pull/414 to improve `obbDisjoint`. Firstly, it has better performance. Secondly, I agree with you that having another implementation for the template specialization with `float` may not be a good idea.

yszheda:
Thank you @sherm1 and @SeanCurtis-TRI !

Before I close this PR, I would like to know if you have any suggestion or idea on how to better exploit the potential of `Eigen3` acceleration.

BTW. Currently there's redundant code in
`template <> bool obbDisjoint(
    const Matrix3<double>& B,
    const Vector3<double>& T,
    const Vector3<double>& a,
    const Vector3<double>& b);`
and
`bool obbDisjoint(
    const Transform3<double>& tf,
    const Vector3<double>& a,
    const Vector3<double>& b);`

would you like another trivial patch to remove the redundancy and add the unit tests?

yszheda:
@SeanCurtis-TRI Sorry for my late reply and thanks for pointing out the problem I haven't noticed before. I agree with you that: 

1. It's not worthwhile just removing the redundant code in `obbDisjoint` and leaving the others.
2. Those interfaces, which take a full transform as one of the parameters, are problematic because they only handle the special cases where transform=rotation+translation.

So I think to refactor this part of code, we need a better patch instead of the superficial way as I suggested before.

SeanCurtis-TRI:
@yszheda Your last comment:

> ...they only handle the special cases where transform=rotation+translation.

particularly caught my attention. I'm curious as to what you meant. Your answer isn't relevant to the closing of this issue. I'm just asking for my own edification. When I read it, it sounds like you're suggesting there are other transforms (e.g., reflections, skews, projections?) that should be considered. Is that what you meant, or did you have something else in mind?

yszheda:
> @yszheda Your last comment:
> When I read it, it sounds like you're suggesting there are other transforms (e.g., reflections, skews, projections?) that should be considered. Is that what you meant, or did you have something else in mind?

Sorry, I misunderstood your previous comment. I thought the problem of those interfaces with a `Tranform3` parameter is that the `Eigen::Transform` type is a more general concept while the implementation only deals with the special case. But after checking the type alias in `types.h`, I find that `Tranform3` is already restricted as isometry.

SeanCurtis-TRI:
No worries.  :) That's why I ask clarifying questions.

Curiously enough, in Drake we're deprecating the use of the `Eigen::Transform` for several reasons:

1. Even in `Isometry` mode, it allows for reflections (which we don't consider to be a valid transformations of physical bodies). It *is* a valid isometry, but that's a larger transformation space than we want.
2. The `Transform` doesn't make any guarantees about the `linear()` basis -- is it properly orthonormal? We've introduced our own `RigidTransform` class that provides guarantees in that regard. We've encountered numerous examples where, for various meaningful reasons, rotations get instantiated where things aren't properly orthonormal which helps accumulate error.

PR #414, opened 2019-07-22
WIP: Optimize obbDisjoint using SIMD intrinsics

yszheda:
Hi,

As mentioned in https://github.com/flexible-collision-library/fcl/pull/413, this is another approach of vectorized obbDisjoint.

Here I just briefly list the results. For more details of the performance comparison experiments, please refer to https://github.com/flexible-collision-library/fcl/pull/413.

* For `fcl::obbDisjoint<float>`:
  * SSE: approximately 1.6x speedup
  * AVX: approximately 1.5x speedup
* For `fcl::obbDisjoint<double>`:
  * SSE: approximately 1.4x speedup
  * AVX: approximately 1.3x speedup

I still have some questions:
* We have deprecated the previous SIMD implementation and decided to rely on `Eigen` for faster math operations. Is it OK for me to add back SIMD intrinsics in `fcl`?
* I have added `FCL_USE_SSE4` and `FCL_USE_AVX` CMake options in a similar way as `FCL_USE_X64_SSE` (BTW. I skipped the Windows version). But I'm wondering if there's better way for the user to specify the CPU capabilities. Perhaps like specifying the CPU capabilities in `make TARGET=`?

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/414)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,12 +42,12 @@ script:
   - cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DFCL_TREAT_WARNINGS_AS_ERRORS=ON -DFCL_COVERALLS=$COVERALLS ..
 
   # Build
-  - make -j4
+  - travis_wait make -j $(nproc)
   - if [ $COVERALLS = ON ] && [ "$TRAVIS_OS_NAME" = "linux" ]; then make coveralls; fi
 
   # Run unit tests
-  - ctest -j4 --output-on-failure
+  - travis_wait ctest -j $(nproc) --output-on-failure
 
   # Make sure we can install and uninstall with no issues
-  - sudo make -j4 install
-  - sudo make -j4 uninstall
+  - sudo make -j $(nproc) install
+  - sudo make -j $(nproc) uninstall
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,7 +87,27 @@ endif()
 
 # Whether to enable SSE
 set(SSE_FLAGS "")
+option(FCL_USE_AVX "Whether FCL should use AVX instructions" OFF)
+option(FCL_USE_SSE4 "Whether FCL should use SSE4 instructions" OFF)
 option(FCL_USE_X64_SSE "Whether FCL should x64 SSE instructions" ON)
+
+if(FCL_USE_AVX)
+  set(FCL_USE_SSE4 OFF)
+  set(FCL_USE_X64_SSE OFF)
+  if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    set(AVX_FLAGS -mavx)
+  endif()
+  add_compile_options(${AVX_FLAGS})
+endif()
+
+if(FCL_USE_SSE4)
+  set(FCL_USE_X64_SSE OFF)
+  if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    set(SSE_FLAGS -mfpmath=sse -msse4)
+  endif()
+  add_compile_options(${SSE_FLAGS})
+endif()
+
 if(FCL_USE_X64_SSE)
   if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     set(SSE_FLAGS -mfpmath=sse -msse -msse2 -msse3 -mssse3)
--- a/include/fcl/math/bv/OBB-inl.h
+++ b/include/fcl/math/bv/OBB-inl.h
@@ -42,8 +42,14 @@
 
 #include "fcl/common/unused.h"
 
+#include "fcl/math/math_simd_details.h"
+
 namespace fcl
 {
+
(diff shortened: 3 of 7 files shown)


yszheda:
Hi @sherm1,

> The comment above ("We ... decided to rely on Eigen for faster math operations.") seems inconsistent with the code below which appears to rely entirely on explicit Intel intrinsics. Please clarify.

Sorry for the confusing words. Here I mean the FCL community decided to replace SIMD intrinsics with Eigen years ago. But my patch uses SIMD intrinsics to achieve better performance of `obbDisjoint`. So I'm wondering if it violates the previous decision by the community.

> The lack of unit tests is particularly worrying for code this complicated!

Thanks for the suggestion. I can provide unit tests later, since I have lots of testcases to choose ;)

yszheda:
> Thanks for the clarification. Have you tried enabling Eigen's SIMD support to compare that with your hand-writting SIMD code?

I forgot to mention that I already turned on `FCL_USE_X64_SSE` (which uses `-mfpmath=sse -msse -msse2 -msse3 -mssse3`) when I tested the performance of the original implementation and the Eigen patch https://github.com/flexible-collision-library/fcl/pull/413. According to [the Eigen FAQ](http://eigen.tuxfamily.org/index.php?title=FAQ#Vectorization), Eigen's SIMD support is supposed to be enabled.

yszheda:
Hi @sherm1 ,

It seems that some Travis CI jobs fail because the compilation time is longer than the default Travis waiting time. Could you please modify the `travis_wait` settings (https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received) or just retry the failed jobs? Thanks!

---

Here is the error log from https://travis-ci.org/flexible-collision-library/fcl/jobs/562447511

```
[ 78%] Building CXX object test/CMakeFiles/test_fcl_broadphase_collision_1.dir/test_fcl_broadphase_collision_1.cpp.o

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
The build has been terminated
```

sherm1:
Trying a build restart.

yszheda:
Hi @sherm1 ,

> Trying a build restart.

Thank you! But the same issue still exists (https://travis-ci.org/flexible-collision-library/fcl/jobs/562447514):

```
27/37 Test  #5: test_fcl_broadphase_distance ....................   Passed  107.89 sec
      Start 31: test_fcl_octomap_distance
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
The build has been terminated
```

> The lack of unit tests is particularly worrying for code this complicated!

I have added unit tests both for the math functions using SIMD intrinsics and for `obbDisjoint`.

sherm1:
Hi, @yszheda. The Travis build is controlled by file `.travis.yml` in the FCL top directory. I believe you can modify that yourself as part of this PR to see what is needed to get it to build. Can you try that and let me know if it doesn't work? Then we can review the build change along with the rest of the PR.

Any thoughts on why the build time went up?

yszheda:
Thanks @sherm1 !

> Any thoughts on why the build time went up?

I've found that there're only 2 CPU cores in the Travis CI environment. However, currently `.travis.yml` in the master branch is using `make -j4` and `ctest -j4` in the script, which will cause thread oversubscription. I'm trying to use `$(nproc)` to replace the hard coded number 4, but it seems that it's not supported in Mac OS X.

yszheda:
Hi @sherm1 ,
I agree with you that it would be better to break this PR into several independent ones. And I will do that after we make sure we have no more way to use Eigen's vectorization more effectively than https://github.com/flexible-collision-library/fcl/pull/413

sherm1:
@yszheda FYI Vectorclass, version2  is another possibility for adding SIMD support:
- [announcement](https://www.agner.org/optimize/blog/read.php?i=1013)
- [GitHub repo](https://github.com/vectorclass/version2)
- [documentation](https://github.com/vectorclass/version2)

We haven't tried this yet -- the open source version was just released. I'm not sure yet how this compares with Eigen's SIMD support, but I suspect it is much better.

yszheda:
Thank you for reviewing the code and providing advice.

* As for the unit tests, I can use fixed testcases rather than random ones. But for a general math function like matrix multiplication, it's hard to provide correctness assurance even if all the fixed tests are passed. In this case, using massive amounts of random testcases is better to avoid bias of specified testcases. But as @SeanCurtis-TRI suggests, a good fuzzy testing framework is needed to capture the failures.

* > We agree that withdrawing this PR is probably best; it's a very ambitious effort and that brings its own challenges in testing and infrastructure.

Yep, I agree.

* > Out of curiosity, how would you characterize your group's efforts?

Oh I haven't list the contribution of my colleagues, simply because only I'm involved in this task (improving the performance of `fcl` obb overlap checking). XD

* For the `vectorclass/version2` library, I'm very willing to have a look at the implementation and try it on the `obbDisjoint` case.

PR #420, opened 2019-11-15
Fix capsule moment of inertia

gabrielebndn:
The formula for `ix` in the capsule was wrong, I fixed it.
Additionally, I grabbed the occasion to group some common terms in the other expressions too

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/420)

### DIFF
--- a/include/fcl/geometry/shape/capsule-inl.h
+++ b/include/fcl/geometry/shape/capsule-inl.h
@@ -82,6 +82,7 @@ S Capsule<S>::computeVolume() const
 }
 
 //==============================================================================
+// Compare https://www.gamedev.net/articles/programming/math-and-physics/capsule-inertia-tensor-r3856/
 template <typename S>
 Matrix3<S> Capsule<S>::computeMomentofInertia() const
 {
--- a/include/fcl/geometry/shape/capsule-inl.h
+++ b/include/fcl/geometry/shape/capsule-inl.h
@@ -85,11 +85,14 @@ S Capsule<S>::computeVolume() const
 template <typename S>
 Matrix3<S> Capsule<S>::computeMomentofInertia() const
 {
-  S v_cyl = radius * radius * lz * constants<S>::pi();
-  S v_sph = radius * radius * radius * constants<S>::pi() * 4 / 3.0;
+  S l2 = lz * lz;
+  S r2 = radius * radius;
 
-  S ix = v_cyl * lz * lz / 12.0 + 0.25 * v_cyl * radius + 0.4 * v_sph * radius * radius + 0.25 * v_sph * lz * lz;
-  S iz = (0.5 * v_cyl + 0.4 * v_sph) * radius * radius;
+  S v_cyl = r2 * lz * constants<S>::pi();
+  S v_sph = r2 * radius * constants<S>::pi() * 4 / 3.0;
+
+  S ix = v_cyl * (l2 / 12. + r2 / 4.) + v_sph * (0.4 * r2 + 0.25 * l2 + 3. * radius * lz / 8.);
+  S iz = (0.5 * v_cyl + 0.4 * v_sph) * r2;
 
   return Vector3<S>(ix, ix, iz).asDiagonal();
 }
--- a/test/geometry/shape/CMakeLists.txt
+++ b/test/geometry/shape/CMakeLists.txt
@@ -1,5 +1,6 @@
 set(tests
         test_convex.cpp
+        test_capsule.cpp
         )
 
 # Build all the tests
--- a/test/geometry/shape/test_capsule.cpp
+++ b/test/geometry/shape/test_capsule.cpp
@@ -56,7 +56,7 @@ std::vector<paird> get_test_sizes() {
       paird(20.,30.),
       paird(3.,2.),
       paird(30.,20.)
-      };
+  };
 }
 
 template <typename S>
@@ -70,7 +70,7 @@ void testVolumeComputation(const Capsule<S>& shape, S tol) {
   S v_sph = 4./3.*pi*r*r*r; // volume of a sphere
   S v_cap = v_cyl + v_sph;  // total volume
 
-  EXPECT_NEAR(shape.computeVolume(), v_cap, tol);
+  EXPECT_TRUE(Eigen::internal::isApprox(shape.computeVolume(), v_cap, tol));
 }
 
 template <typename S>
@@ -113,12 +113,12 @@ GTEST_TEST(Capsule, Volume_Capsule) {
     double rd = pair.first;
     double ld = pair.second;
     Capsuled capsule_d(rd, ld);
-    testVolumeComputation(capsule_d, 1e-10);
+    testVolumeComputation(capsule_d, 1e-15);
 
     double rf = static_cast<float>(pair.first);
     double lf = static_cast<float>(pair.second);
     Capsulef capsule_f(rf, lf);
-    testVolumeComputation(capsule_f, 1e-9f);
+    testVolumeComputation(capsule_f, 1e-8f);
   }
 }
 
--- a/test/geometry/shape/test_capsule.cpp
+++ b/test/geometry/shape/test_capsule.cpp
@@ -32,11 +32,11 @@
  *  POSSIBILITY OF SUCH DAMAGE.
  */
 
-/** @author Sean Curtis (sean@tri.global) (2018) */
+/** @author Gabriele Buondonno (gbuondon@laas.fr) (2019) */
 
-// Tests the implementation of a convex polytope geometry.
+// Tests the implementation of a capsule object.
 
-#include "fcl/geometry/shape/convex.h"
+#include "fcl/geometry/shape/capsule.h"
 
 #include <vector>
 
@@ -44,7 +44,6 @@
 #include <gtest/gtest.h>
 
 #include "eigen_matrix_compare.h"
-#include "fcl/geometry/shape/capsule.h"
 
 namespace fcl {
 namespace {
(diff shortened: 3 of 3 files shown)


tehbelinda:
:lgtm:

tehbelinda:
___
*[include/fcl/geometry/shape/capsule-inl.h, line 94 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/420#-Lu3iQ2xBXEt9k_eIc91:-Lu3iQ2xBXEt9k_eIc92:b-n7b6yb) ([raw file](https://github.com/flexible-collision-library/fcl/blob/100781bd6d0f064a940dce077c13b6482c713bd0/include/fcl/geometry/shape/capsule-inl.h#L94)):*
> ```objc
>   S v_sph = r2 * radius * constants<S>::pi() * 4 / 3.0;
> 
>   S ix = v_cyl * (l2 / 12. + r2 / 4.) + v_sph * (0.4 * r2 + 0.25 * l2 + 3. * radius * lz / 8.);
> ```

could you also add a comment with a link to your source for this?

sherm1:
@gabrielebndn thanks -- this is an improvement since it was broken before. However, one might wonder how it could have been wrong for so long. IMO that is because there was no unit testing and that remains true with the fix you are proposing. We have been trying to upgrade FCL's software quality as we make changes to it. GTEST is easily usable and you can find a test case we added in fcl/test/geometry/shape/test_convex.cpp. Please consider adding a short test_capsule.cpp in that same directory that verifies correct capsule inertias for a few representative cases where you can compute the correct answers by hand.

Please note that the main benefit of unit tests is not to check the code you just wrote (which I trust you engineered carefully and Bel verified) but to make sure that the code continues to work correctly forever in the face of unknown future naive programmers who might think they can improve on your work. It also provides proof to future skeptics that your code is correct.

gabrielebndn:
@tehbelinda, I could not find authoritative sources.
Verification was prompted, among other things, by the fact that I noticed the term `0.25 * v_cyl * radius` was clearly incorrect, since dimensionally it did not make sense.
I followed the approach at https://www.gamedev.net/articles/programming/math-and-physics/capsule-inertia-tensor-r3856/  and I personally verified the calculations by hand. To complicate things, notice Eq. (14) contains a typo: the term H^2/2 is not consistent with Eq. (12) or with the code below.
Is it ok as a reference? Should I add this link as a comment to the code?

@sherm1, I fully understand the necessity of unit tests. I will try to provide one.

tehbelinda:
thanks @gabrielebndn, that link would be great

gabrielebndn:
I pushed a unit test.
1. Some tests are quite obvious. In particular, the CoM is just checking that it is zero and the volume check is almost a copy of the actual method
1. To compute the moment of inertia, I am basically performing the computation inside the unit test, assuming the formula for the inertia of a sphere, formula for the inertia of a cylinder, and the CoM of a hemisphere
1. To compare moments of inertia, I am using Eigen's `isApprox` instead of `CompareMatrices`: I was getting differences of a few units for large inertia values (order of 10^5). `isApprox` measures difference by percent, which is more important, I think. It's the method I've always used.
1. Something funny is going on with the floating point literals. If I write `S(4./3.)` or `S(4.)/S(3.)` instead of `4./3.`at line 71 of the unit test, it does not pass. The fact is floating point literals are not converted to the correct scalar type inside the code. So this is what happens for instance in https://github.com/flexible-collision-library/fcl/blob/08d0bb25c290a4a5e713851d2a5ac514e4355801/include/fcl/geometry/shape/capsule-inl.h#L81
   1. The literals are interpreted as `double`
   1. `float` variables, since they are operating with `double`, are promoted to `double`
   1. Operations are carried out as `double`
   1. The final result is implicitly converted back to `float`
If this behavior is unwanted then the whole code should be verified so that each floating point literal is explicitly converted to the correct scalar type.
1. I added a comment to `computeMomentofInertia` including the link I posted above

gabrielebndn:
> _[test/geometry/shape/test_capsule.cpp, line 122 at r2](https://reviewable.io/reviews/flexible-collision-library/fcl/420#-Lu5UZYP3XGFI9Ab9Im5:-Lu5UZYP3XGFI9Ab9Im6:b-an55v4) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4a4b978b63aab76c5cdacca405f52d1b7ff72fa6/test/geometry/shape/test_capsule.cpp#L122)):_
> 
> > ```c++
> >     double lf = static_cast<float>(pair.second);
> >     Capsulef capsule_f(rf, lf);
> >     testVolumeComputation(capsule_f, 1e-9f);
> > ```
> 
> If you can get this level of accuracy with float precision, you should be able to do much better in double. So the double test tolerance should be much tighter, maybe 1e-14 ish? BTW isn't 1e-9 a little extreme for float precision? -- I'm surprised it matches that closely. Consider using a small multiple of `numeric_limits<S>::epsilon()` instead to make sure the test is robust.

For the test with double precision, I manage to get down to 1e-10.
The fact I am able to reach a good accuracy with float precision might be due to what I explain above: I think that, due to incorrect handling of the floating point literals, computations are actually performed in double precision.
I am unfamiliar with floats, as I always use doubles, so I have no idea of how much error to expect. Should I cast the literals so that all computations are carried out in float precision? This will likely reduce accuracy a lot.

> _[test/geometry/shape/test_capsule.cpp, line 147 at r2](https://reviewable.io/reviews/flexible-collision-library/fcl/420#-Lu5VmuW13bxWS9iUUSI:-Lu5VmuW13bxWS9iUUSJ:b4kye5u) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4a4b978b63aab76c5cdacca405f52d1b7ff72fa6/test/geometry/shape/test_capsule.cpp#L147)):_
> 
> > ```c++
> >     double ld = pair.second;
> >     Capsuled capsule_d(rd, ld);
> >     testMomentOfInertiaComputation(capsule_d, 1e-6);
> > ```
> 
> That seems very loose for a double precision test.

Indeed, I manage to bring this to 1e-14. I remind that this is the tolerance sent to `isApprox`.

jmirabel:
> Howevever, the 1e-10 error in double is worrisome. That is a six orders-of-magnitude loss from the underlying precision of 2e-16 ish. For such a simple computation, that would only make sense if this is an absolute tolerance applied to a large-scale problem (like the actual volume is 10⁶ ish. If that's the case it would be better to change the testVolumeComputation() method to accept a relative tolerance rather than absolute so that it is clear that you are not suffering a dramatic precision loss somewhere.

You very easily reach 10^6 with a formula with has terms like radius^3 * length... The unit test has radius = 30 and length = 20.

gabrielebndn:
> Looks good to me with one requested change, please take a look (PTAL).
> 
> > Reviewed 1 of 1 files at r3.
> > _[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/420#-:-Lu8e77T5Yvqox-rL4zJ:b-or07ri)_ status: all files reviewed, 2 unresolved discussions (waiting on @gabrielebndn)
> 
> _[test/geometry/shape/test_capsule.cpp, line 122 at r2](https://reviewable.io/reviews/flexible-collision-library/fcl/420#-Lu5UZYP3XGFI9Ab9Im5:-Lu8cPS3DwrQftOYEGud:b-r2e27s) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4a4b978b63aab76c5cdacca405f52d1b7ff72fa6/test/geometry/shape/test_capsule.cpp#L122)):_
> 
> Previously, gabrielebndn wrote…
> > For the test with double precision, I manage to get down to 1e-10. The fact I am able to reach a good accuracy with float precision might be due to what I explain above: I think that, due to the fact floating point literals are not cast to the templated scalar type, computations are actually performed in double precision. I am unfamiliar with floats, as I always use doubles, so I have no idea of how much error to expect. Should I cast the literals so that all computations are carried out in float precision? This will likely reduce accuracy a lot. Is there any circumstance under which this change may be desirable?
> 
> I'm not worried about the `float` extra precision -- like you I only use `double` and using `float` rarely provides the performance improvements that programmers often think it does.
> 
> Howevever, the 1e-10 error in double is worrisome. That is a six orders-of-magnitude loss from the underlying precision of 2e-16 ish. For such a simple computation, that would only make sense if this is an absolute tolerance applied to a large-scale problem (like the actual volume is 10⁶ ish. If that's the case it would be better to change the testVolumeComputation() method to accept a relative tolerance rather than absolute so that it is clear that you are not suffering a dramatic precision loss somewhere.

As @jmirabel points out, the sizes are indeed quite big. The one test that does not pass 1e-11 involves a volume of about 71209 m^3, for all other I can get to 1e-13 (after this threshold there is one more test, 71.209 m^3, which barely exceeds 1e-14).

I set up the tests so that several different sizes are employed using test_convex as a blueprint, as I thought it wise to test several different values in order to avoid that the test passes only because of a coincidental choice of the parameters; to achieve this goal, I might as well have chosen numbers in the same order of magnitude. I did not think I would incur in such problems with the scaling.
However, I do think that relative error makes more sense when dealing with floating point values. Therefore, if you like this solution, I was happy to implement it.

gabrielebndn:
> minor: I would be more comfortable with something more like 1e-8f, which is all we can expect from float precision. The fact that these computations are being done secretly in double provides more accuracy, but I don't think we should insist on that in the test (it would be nice if the test could still pass if someone sets compiler flags properly to force computations to actually use float).

Agreed, I was wondering which choice would be more accurate

gabrielebndn:
Yes, I saw the failing test. I am inclined to think it is unrelated to my change, or at least not *directly* related.
That test, `test_fcl_signed_distance`, seems to only perform distance computations. I would be surprised if they involved the moments of inertia.
If I interpret the results correctly, the failing unit test is `RealWorldRegression`: this test does not seem to even involve capsules at all, only boxes.
If you look at the historical record of the CI builds, my first PR was passing. Then they are failing, but the only changes were to the unit test and a comment in the main code. It is always `RealWorldRegression` in `test_fcl_signed_distance` which is failing, but not only on Mac: Build 1392 fails on a Linux, 1393 on a Linux *and* on Mac, 1395 on Mac only.
My guess is that the error appears randomly because of something like uninitialized memory involved in that test: either in the routines, or in the unit test itself.

gabrielebndn:
> minor: } is over indented

Fixed :)

gabrielebndn:
Well, it seems Travis *is* passing right now. Somebody should really take a look into `test_fcl_signed_distance`. But as far as this PR is concerned, everything seems fine.

PR #423, opened 2019-11-25
Add capsule test for local AABB computation

tehbelinda:
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/423)

### DIFF
--- a/test/geometry/shape/test_capsule.cpp
+++ b/test/geometry/shape/test_capsule.cpp
@@ -60,6 +60,16 @@ std::vector<paird> get_test_sizes() {
 }
 
 template <typename S>
+void testLocalAABBComputation(Capsule<S>& shape, S tol) {
+  shape.computeLocalAABB();
+
+  S r = shape.radius;
+  S l = shape.lz;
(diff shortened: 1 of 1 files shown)


tehbelinda:
___
*[test/geometry/shape/test_capsule.cpp, line 127 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/423#-LuZ1POH13pvi1fRyUuC:-LuZ3BkY59u5Vv7k8O_o:b-1wys3b) ([raw file](https://github.com/flexible-collision-library/fcl/blob/2b02130ec72722e67f36b184a8c745cd237d1483/test/geometry/shape/test_capsule.cpp#L127)):*

nit `rf` and `lf` should be floats.

As an alternative, if you put each test in its own block, you could re-use the variable names, merely changing the type.
</blockquote></details>

I was trying to keep it consistent with the rest of the file, though I see those are all doubles so I'll change them to floats too

tehbelinda:
looks like the same flakey test as the last PR :S

tehbelinda:
Ok thanks

Issue #407, opened 2019-06-09
Segmentation fault on Box-Cylinder signed distance (ccd-GJK)


wxmerkt:
Using  the latest master (2112037d3f0490f83c9e2d237886eb113b7b0d31), I get the following kind of segmentation faults _fairly_ often:

```
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x00007fffa534a313 in fcl::detail::libccd_extension::ccdGJKSignedDist(void const*, void const*, _ccd_t const*, _ccd_vec3_t*, _ccd_vec3_t*) () from [path]/lib/libfcl.so.0.6
(gdb) bt
#0  0x00007fffa534a313 in fcl::detail::libccd_extension::ccdGJKSignedDist(void const*, void const*, _ccd_t const*, _ccd_vec3_t*, _ccd_vec3_t*) () from [path]/lib/libfcl.so.0.6
#1  0x00007fffa5348931 in bool fcl::detail::GJKSignedDistance<double>(void*, void (*)(void const*, _ccd_vec3_t const*, _ccd_vec3_t*), void*, void (*)(void const*, _ccd_vec3_t const*, _ccd_vec3_t*), unsigned int, double, double*, Eigen::Matrix<double, 3, 1, 0, 3, 1>*, Eigen::Matrix<double, 3, 1, 0, 3, 1>*) ()
   from [path]/lib/libfcl.so.0.6
#2  0x00007fffa532c805 in fcl::detail::ShapeDistanceTraversalNode<fcl::Cylinder<double>, fcl::Box<double>, fcl::detail::GJKSolver_libccd<double> >::leafTesting(int, int) const () from [path]/lib/libfcl.so.0.6
#3  0x00007fffa536bf2f in void fcl::detail::distance<double>(fcl::detail::DistanceTraversalNodeBase<double>*, std::__cxx11::list<fcl::detail::BVHFrontNode, std::allocator<fcl::detail::BVHFrontNode> >*, int) ()
   from [path]/lib/libfcl.so.0.6
[...]

PR #427, opened 2019-11-28
fix missing nearest points in Octree<->Mesh distance result

SebastianRiedel:
I stumbled upon this one while testing distance queries between octrees and other entities. Seems like someone just forgot to add the (already computed) nearest points to the result object in the Octree<->Mesh case. A quick test suggests that the nearest points returned now are correct.

For comparison, the Octree<->[Shape, Octree] distance computation attaches the nearest points to the result, cf. https://github.com/flexible-collision-library/fcl/blob/6aeb7a1d30ae57b6b5413b65e0bf2ce9800687ed/include/fcl/narrowphase/detail/traversal/octree/octree_solver-inl.h#L297 and https://github.com/flexible-collision-library/fcl/blob/6aeb7a1d30ae57b6b5413b65e0bf2ce9800687ed/include/fcl/narrowphase/detail/traversal/octree/octree_solver-inl.h#L787

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/427)

### DIFF
--- a/include/fcl/narrowphase/detail/traversal/octree/octree_solver-inl.h
+++ b/include/fcl/narrowphase/detail/traversal/octree/octree_solver-inl.h
@@ -506,7 +506,7 @@ bool OcTreeSolver<NarrowPhaseSolver>::OcTreeMeshDistanceRecurse(const OcTree<S>*
       Vector3<S> closest_p1, closest_p2;
       solver->shapeTriangleDistance(box, box_tf, p1, p2, p3, tf2, &dist, &closest_p1, &closest_p2);
 
-      dresult->update(dist, tree1, tree2, root1 - tree1->getRoot(), primitive_id);
+      dresult->update(dist, tree1, tree2, root1 - tree1->getRoot(), primitive_id, closest_p1, closest_p2);
 
       return drequest->isSatisfied(*dresult);
     }
--- a/test/test_fcl_octomap_distance.cpp
+++ b/test/test_fcl_octomap_distance.cpp
@@ -65,6 +65,8 @@ void octomap_distance_test_BVH(std::size_t n, double resolution = 0.1);
 template <typename S>
 void test_octomap_distance()
 {
+  // -- compares result of (manager of primitives <-> octomap)
+  //    against (manager of primitives <-> manager of box primitives from octomap)
 #ifdef NDEBUG
   octomap_distance_test<S>(200, 100, false, false);
   octomap_distance_test<S>(200, 1000, false, false);
@@ -83,6 +85,8 @@ GTEST_TEST(FCL_OCTOMAP, test_octomap_distance)
 template <typename S>
 void test_octomap_distance_mesh()
 {
+  // -- compares result of (manager of meshes <-> octomap)
+  //    against (manager of meshes <-> manager of box meshes from octomap)
 #ifdef NDEBUG
   octomap_distance_test<S>(200, 100, true, true);
   octomap_distance_test<S>(200, 1000, true, true);
@@ -101,6 +105,8 @@ GTEST_TEST(FCL_OCTOMAP, test_octomap_distance_mesh)
 template <typename S>
 void test_octomap_distance_mesh_octomap_box()
 {
+  // -- compares result of (manager of meshes <-> octomap)
+  //    against (manager of meshes <-> manager of box primitives from octomap)
 #ifdef NDEBUG
   octomap_distance_test<S>(200, 100, true, false);
   octomap_distance_test<S>(200, 1000, true, false);
@@ -219,6 +225,13 @@ void octomap_distance_test_BVH(std::size_t n, double resolution)
 
     std::cout << dist1 << " " << dist2 << std::endl;
     EXPECT_TRUE(std::abs(dist1 - dist2) < 0.001);
+
+    std::cout << "cdata, np0:\n" << cdata.result.nearest_points[0] << std::endl;
+    std::cout << "cdata, np1:\n" << cdata.result.nearest_points[1] << std::endl;
+    std::cout << "cdata2, np0:\n" << cdata2.result.nearest_points[0] << std::endl;
+    std::cout << "cdata2, np1:\n" << cdata2.result.nearest_points[1] << std::endl;
+    EXPECT_NEAR((cdata.result.nearest_points[0] - cdata2.result.nearest_points[1]).norm(), 0.0, 1e-6);
+    EXPECT_NEAR((cdata.result.nearest_points[1] - cdata2.result.nearest_points[0]).norm(), 0.0, 1e-6);
   }
 }
 
@@ -293,6 +306,13 @@ void octomap_distance_test(S env_scale, std::size_t env_size, bool use_mesh, boo
     EXPECT_TRUE(cdata2.result.min_distance <= 0);
   else
     EXPECT_TRUE(std::abs(cdata.result.min_distance - cdata2.result.min_distance) < 1e-3);
+
+  std::cout << "cdata, np0:\n" << cdata.result.nearest_points[0] << std::endl;
+  std::cout << "cdata, np1:\n" << cdata.result.nearest_points[1] << std::endl;
+  std::cout << "cdata2, np0:\n" << cdata2.result.nearest_points[0] << std::endl;
+  std::cout << "cdata2, np1:\n" << cdata2.result.nearest_points[1] << std::endl;
+  EXPECT_NEAR((cdata.result.nearest_points[0] - cdata2.result.nearest_points[0]).norm(), 0.0, 1e-6);
+  EXPECT_NEAR((cdata.result.nearest_points[1] - cdata2.result.nearest_points[1]).norm(), 0.0, 1e-6);
 
   delete manager;
   delete manager2;
--- a/test/test_fcl_octomap_distance.cpp
+++ b/test/test_fcl_octomap_distance.cpp
@@ -65,8 +65,6 @@ void octomap_distance_test_BVH(std::size_t n, double resolution = 0.1);
 template <typename S>
 void test_octomap_distance()
 {
-  // -- compares result of (manager of primitives <-> octomap)
-  //    against (manager of primitives <-> manager of box primitives from octomap)
 #ifdef NDEBUG
   octomap_distance_test<S>(200, 100, false, false);
   octomap_distance_test<S>(200, 1000, false, false);
@@ -85,8 +83,6 @@ GTEST_TEST(FCL_OCTOMAP, test_octomap_distance)
 template <typename S>
 void test_octomap_distance_mesh()
 {
-  // -- compares result of (manager of meshes <-> octomap)
-  //    against (manager of meshes <-> manager of box meshes from octomap)
 #ifdef NDEBUG
   octomap_distance_test<S>(200, 100, true, true);
   octomap_distance_test<S>(200, 1000, true, true);
@@ -105,8 +101,6 @@ GTEST_TEST(FCL_OCTOMAP, test_octomap_distance_mesh)
 template <typename S>
 void test_octomap_distance_mesh_octomap_box()
 {
-  // -- compares result of (manager of meshes <-> octomap)
-  //    against (manager of meshes <-> manager of box primitives from octomap)
 #ifdef NDEBUG
   octomap_distance_test<S>(200, 100, true, false);
   octomap_distance_test<S>(200, 1000, true, false);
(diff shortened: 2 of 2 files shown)


AlbertoMarnetto-Work:
As I need this functionality in a project of mine, I would like to thank the author for finding and fixing this error.

I shortly tried the fix with a test case of mine. After applying the patch the `distanceResult.nearest_points` array is filled, however I noticed that the order of the points does not always correspond to the order of the collision objects (i.e. sometimes `nearest_points[0]` belongs to `o2` and `nearest_points[1]` to `o1`).

After further analysis it seems that `nearest_points[0]` belongs to the OcTree object and `nearest_points[1]` to the Mesh, independently of the order of such objects in the call to `distance`. However I did not yet test extensively enough to check if this is always the case (I can provide my sample test if needed).

Is this an intentional behavior, a bug or there is simply no guarantee about the order of `distanceResult.nearest_points`?

jmirabel:
> however I noticed that the order of the points does not always correspond to the order of the collision objects (i.e. sometimes nearest_points[0] belongs to o2 and nearest_points[1] to o1).

This is a bug coming from https://github.com/flexible-collision-library/fcl/blob/6aeb7a1d30ae57b6b5413b65e0bf2ce9800687ed/include/fcl/narrowphase/distance-inl.h#L117 for distance and from https://github.com/flexible-collision-library/fcl/blob/6aeb7a1d30ae57b6b5413b65e0bf2ce9800687ed/include/fcl/narrowphase/collision-inl.h#L123 for collision.

When the test is true, the test that is done is o2, o1 and the result should be inverted. Applying something similar to https://github.com/humanoid-path-planner/hpp-fcl/commit/177fb1fab82cdb27435d9ca590d14b6169235730 should solve the issue.

jmirabel:
> This PR is about octree, but the code you linked to seems to be about geom vs bvh. There is an "object type" of type OT_OCTREE, I would assume that shape-octtree queries would report types of OT_GEOM and OT_OCTREE. If so, this particular test wouldn't affect it. Is that not the case?

You are right. For OcTree <-> BVH, the invertion of the result should happen here: https://github.com/flexible-collision-library/fcl/blob/9dba579158109c0164bfe0e8b4a75c16bfc788f6/include/fcl/narrowphase/detail/traversal/octree/octree_solver-inl.h#L153

AlbertoMarnetto-Work:
> @AlbertoMarnetto-Work Still, while may not be a reasonable expectation for which object is o1 and which is o2, there is a _very_ reasonable expectation that the first nearest point should _always_ correspond to o1 and the second with o2.
> 
> You don't happen to have a code snippet that can reproduce this error, do you? That'd go great in an issue.
> 

Sure, here is a minimal self-contained example, which could hopefully be recycled in a unit test.

```c++
#include <boost/format.hpp>
#include <fcl/math/bv/OBBRSS.h>
#include <fcl/math/triangle.h>
#include <fcl/geometry/bvh/BVH_model.h>
#include <fcl/narrowphase/collision.h>
#include <fcl/narrowphase/distance.h>
#include <Eigen/Core>

#include <vector>

std::string format_vector3(const Eigen::Vector3d& vector)
{
	return str(boost::format("[ %5.2f , %5.2f , %5.2f ]")
			% vector.x() % vector.y() % vector.z());
}

int main()
{
	// Part 1: build a tetrahedron as BVHModel
	std::vector<Eigen::Vector3d> vertices
		= { { 0, 0, 0 }, { 1, 0, 0 }, { 0, 1, 0 }, { 0, 0, 1 } };
	std::vector<fcl::Triangle> triangles
		= { { 0, 1, 2 }, { 0, 1, 3 }, { 0, 2, 3 }, { 1, 2, 3 } };

	typedef fcl::BVHModel<fcl::OBBRSSd> BvhGeometry;
	auto bvhGeometry = std::make_shared<BvhGeometry>();
	bvhGeometry->beginModel();
	bvhGeometry->addSubModel(vertices, triangles);
	bvhGeometry->endModel();

	const fcl::CollisionObjectd bvhCollisionObject(bvhGeometry);

	// Part 2: build a tetrahedron as octree
	auto ocTree = std::make_shared<octomap::OcTree>(/* resolution = */ 0.01);
	for (double x = 0.; x < 1.01; x += 0.05)
	{
		for (double y = 0.; y < 1.01; y += 0.05)
		{
			for (double z = 0.; z < 1.01; z += 0.05)
			{
				bool isOccupied = (x + y + z < 1);
				ocTree->updateNode(octomap::point3d(x, y, z), isOccupied);
			}
		}
	}
	auto ocTreeGeometry = std::make_shared<fcl::OcTreed>(ocTree);
	fcl::CollisionObjectd ocTreeCollisionObject(ocTreeGeometry);

	const fcl::Transform3d xTranslationMatrix(
			Eigen::Translation3d(5., 0., 0.));
	ocTreeCollisionObject.setTransform(xTranslationMatrix);

	// Part 3: perform a distance request where o1 = OcTree and o2 = BVHModel

	const fcl::DistanceRequestd distanceRequest(
			/* enable_nearest_points = */ true);
	fcl::DistanceResultd distanceResult;

	{
		const double distance
				= fcl::distance(&ocTreeCollisionObject, &bvhCollisionObject, distanceRequest, distanceResult);

		std::cout << str(boost::format("distance : %5.2f ,  ")
				% distance
				);

		std::cout << str(boost::format("p1 : %s ,  p2 : %s , ")
				% format_vector3(distanceResult.nearest_points[0])
				% format_vector3(distanceResult.nearest_points[1])
				);

		std::cout << "\n";
		// Output:
		// distance :  4.00 ,  p1 : [  5.00 ,  0.00 ,  0.00 ] ,  p2 : [  1.00 ,  0.00 ,  0.00 ] ,

	}

	// Part 4: perform a distance request swapping o1 and o2
	// We could expect that the returned nearest points are also swapped,
	// but this is not the case: they are returned in the same order
	{
		const double distance
				= fcl::distance(&bvhCollisionObject, &ocTreeCollisionObject, distanceRequest, distanceResult);

		std::cout << str(boost::format("distance : %5.2f ,  ")
				% distance
				);

		std::cout << str(boost::format("p1 : %s ,  p2 : %s , ")
				% format_vector3(distanceResult.nearest_points[0])
				% format_vector3(distanceResult.nearest_points[1])
				);

		std::cout << "\n";
		// Output:
		// distance :  4.00 ,  p1 : [  5.00 ,  0.00 ,  0.00 ] ,  p2 : [  1.00 ,  0.00 ,  0.00 ] ,
	}
}
```

AlbertoMarnetto-Work:
> @AlbertoMarnetto-Work: As I look at your example, it's suggestive that the distance results are presenting the nearest points in a fixed order. Can you also confirm that the `o1` and `o2` are always presented in a fixed order (i.e., `o1` is always the octree and `o2` the bvh)? It seems to me that there is no _real_ problem as long as the nearest points correctly correspond to the returned objects.
> 

Yes, thanks for the suggestion, I had not thought to check the order of the collision objects in the  `DistanceResult`.

Indeed `o1` is always the OcTree and `o2` the BVH, so that `nearest_points[0]` always corresponds to the returned `o1` and `nearest_points[1]` always corresponds to `o2`.

Taking this element into account all my issues are solved. Thanks for the very prompt feedback!

I post here an amended test which checks the identity of the collision objects.

```c++
#include <boost/format.hpp>
#include <fcl/math/bv/OBBRSS.h>
#include <fcl/math/triangle.h>
#include <fcl/geometry/bvh/BVH_model.h>
#include <fcl/narrowphase/collision.h>
#include <fcl/narrowphase/distance.h>
#include <Eigen/Core>

#include <vector>

std::string format_vector3(const Eigen::Vector3d& vector)
{
	return str(boost::format("[ %5.2f , %5.2f , %5.2f ]")
			% vector.x() % vector.y() % vector.z());
}

int main()
{
	// Part 1: build a tetrahedron as BVHModel
	std::vector<Eigen::Vector3d> vertices
		= { { 0, 0, 0 }, { 1, 0, 0 }, { 0, 1, 0 }, { 0, 0, 1 } };
	std::vector<fcl::Triangle> triangles
		= { { 0, 1, 2 }, { 0, 1, 3 }, { 0, 2, 3 }, { 1, 2, 3 } };

	typedef fcl::BVHModel<fcl::OBBRSSd> BvhGeometry;
	auto bvhGeometry = std::make_shared<BvhGeometry>();
	bvhGeometry->beginModel();
	bvhGeometry->addSubModel(vertices, triangles);
	bvhGeometry->endModel();

	const fcl::CollisionObjectd bvhCollisionObject(bvhGeometry);

	// Part 2: build a tetrahedron as octree
	auto ocTree = std::make_shared<octomap::OcTree>(/* resolution = */ 0.01);
	for (double x = 0.; x < 1.01; x += 0.05)
	{
		for (double y = 0.; y < 1.01; y += 0.05)
		{
			for (double z = 0.; z < 1.01; z += 0.05)
			{
				bool isOccupied = (x + y + z < 1);
				ocTree->updateNode(octomap::point3d(x, y, z), isOccupied);
			}
		}
	}
	auto ocTreeGeometry = std::make_shared<fcl::OcTreed>(ocTree);
	fcl::CollisionObjectd ocTreeCollisionObject(ocTreeGeometry);

	const fcl::Transform3d xTranslationMatrix(
			Eigen::Translation3d(5., 0., 0.));
	ocTreeCollisionObject.setTransform(xTranslationMatrix);

	// Part 3: perform a distance request where tree1 = OcTree and tree2 = BVHModel

	const fcl::DistanceRequestd distanceRequest(
			/* enable_nearest_points = */ true);
	fcl::DistanceResultd distanceResult;

	auto collisionObjectIdentity = [&](const fcl::CollisionGeometryd* o) -> std::string
	{
		if (o == &*bvhCollisionObject.collisionGeometry())
			return "BVH";
		else if (o == &*ocTreeCollisionObject.collisionGeometry())
			return "OcTree";
		else
			return "Bad pointer";
	};

	{
		const double distance
				= fcl::distance(&ocTreeCollisionObject, &bvhCollisionObject, distanceRequest, distanceResult);

		std::cout << str(boost::format("distance : %5.2f ,  ")
				% distance
				);

		std::cout << str(boost::format("p1 : %s (in %s),  p2 : %s (in %s), ")
				% format_vector3(distanceResult.nearest_points[0])
				% collisionObjectIdentity(distanceResult.o1)
				% format_vector3(distanceResult.nearest_points[1])
				% collisionObjectIdentity(distanceResult.o2)
				);

		std::cout << "\n";

		// Output:
		// distance :  4.00 ,  p1 : [  5.00 ,  0.00 ,  0.00 ] (in OcTree),  p2 : [  1.00 ,  0.00 ,  0.00 ] (in BVH)

	}

	// Part 4: perform a distance request swapping tree1 and tree2
	{
		const double distance
				= fcl::distance(&bvhCollisionObject, &ocTreeCollisionObject, distanceRequest, distanceResult);

		std::cout << str(boost::format("distance : %5.2f ,  ")
				% distance
				);

		std::cout << str(boost::format("p1 : %s (in %s),  p2 : %s (in %s), ")
				% format_vector3(distanceResult.nearest_points[0])
				% collisionObjectIdentity(distanceResult.o1)
				% format_vector3(distanceResult.nearest_points[1])
				% collisionObjectIdentity(distanceResult.o2)
				);

		std::cout << "\n";
		// Output (same as in part 3):
		// distance :  4.00 ,  p1 : [  5.00 ,  0.00 ,  0.00 ] (in OcTree),  p2 : [  1.00 ,  0.00 ,  0.00 ] (in BVH)
	}
}
```

SeanCurtis-TRI:
@AlbertoMarnetto-Work Glad to hear at least your problem is resolved.  :)

AlbertoMarnetto-Work:
> The _big_ request is could you add a test that fails without your fix and passes with it? As long as we're cleaning things up, it would be great if we could add some rigor to the code base.

I have created a test for this patch. It can be cherry-picked into @SebastianRiedel 's patch or merged afterwards:

https://github.com/AlbertoMarnetto-Work/fcl/commit/eb1e0b205980781bac1bc2586c7dee2f601235a9

AlbertoMarnetto-Work:
I think the failure is due to the ill-posedness of the problem. It suffices to change the geometry of a small amount (e.g. switching from the octomap to its "boxes" approximation) to make a different pair of points the nearest one.

For this reason I would suggest a simpler check (s. my commit), which even if weaker suffices to prove that the fix is improving the codebase.

AlbertoMarnetto-Work:
Any news on this? I think that the combination @SebastianRiedel 's fix with my test (https://github.com/AlbertoMarnetto-Work/fcl/commit/eb1e0b205980781bac1bc2586c7dee2f601235a9) yields a fully working patch. The test is slightly weaker than Sebastian's one, but it passes and suffices to show that the calculation of the nearest points works.

The branch is already available:

https://github.com/flexible-collision-library/fcl/compare/master...AlbertoMarnetto-Work:fix-octomap-mesh-distance-test

If Sebastian agrees I would open a PR for it.

AlbertoMarnetto:
The CI failed, but this seems to be unrelated with this patch (segfault in `test_fcl_signed_distance`, s. #407).

The test involved in this branch (`test_fcl_octomap_distance`) was instead successful in all builds.

SeanCurtis-TRI:
Merged despite erratic test failure in unrelated test.

PR #441, opened 2020-02-07
Fix illegal preprocessor expression when Octree enabled

sherm1:
When building on Windows with Octree enabled I ran into compilation failures due to non-standard preprocessor expressions.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/441)

### DIFF
--- a/include/fcl/narrowphase/detail/traversal/octree/collision/mesh_octree_collision_traversal_node.h
+++ b/include/fcl/narrowphase/detail/traversal/octree/collision/mesh_octree_collision_traversal_node.h
@@ -39,7 +39,7 @@
 #define FCL_TRAVERSAL_OCTREE_MESHOCTREECOLLISIONTRAVERSALNODE_H
 
 #include "fcl/config.h"
-#if not(FCL_HAVE_OCTOMAP)
+#ifndef FCL_HAVE_OCTOMAP
 #error "This header requires fcl to be compiled with octomap support"
 #endif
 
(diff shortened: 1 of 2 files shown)


include/fcl/narrowphase/detail/traversal/octree/collision/mesh_octree_collision_traversal_node.h
```diff
@@ -39,7 +39,7 @@
 #define FCL_TRAVERSAL_OCTREE_MESHOCTREECOLLISIONTRAVERSALNODE_H
 
 #include "fcl/config.h"
-#if not(FCL_HAVE_OCTOMAP)
+#ifndef FCL_HAVE_OCTOMAP
```
jslee02: `FCL_HAVE_OCTOMAP` will be always defined either `FCL_HAVE_OCTOMAP 0` or `FCL_HAVE_OCTOMAP 1` since we're using [`cmakedefine01`](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/config.h.in#L45) (see [CMake manual](https://cmake.org/cmake/help/v3.0/command/configure_file.html)).

Maybe `#if !(FCL_HAVE_OCTOMAP)` would work instead? (https://docs.microsoft.com/en-us/cpp/preprocessor/hash-if-hash-elif-hash-else-and-hash-endif-directives-c-cpp?view=vs-2019)

jslee02:
My best guess is that MSVC interprets `not(0)` as `false` (since octomap is not installed on Appveyor), which I don't know why. 🤔

sherm1:
VS 2019 didn't like `#if not(XXX)` at all -- it said "Illegal token after preprocessor directive.".

jslee02:
Ah, I see. Appveyor CI test was passed because these headers were never included so the condition was not even checked (e.g., [this](https://github.com/flexible-collision-library/fcl/blob/9dba579158109c0164bfe0e8b4a75c16bfc788f6/include/fcl/broadphase/broadphase_dynamic_AABB_tree_array-inl.h#L43-L45)).

Issue #458, opened 2020-04-04
Shared lib support is broken on Windows


SpaceIm:
- fcl fails to build as a shared lib whatever Visual Studio version is used.
- fcl can be build as a shared lib with MinGW (when it doesn't fail du to big obj), bit it fails to be consumed.

I suspect a bad use of dll_export/dll_import in front of templates.

Here is a log for MinGW:

[fcl_mingw_shared.txt](https://github.com/flexible-collision-library/fcl/files/4431082/fcl_mingw_shared.txt)

PR #461, opened 2020-04-07
fix shared on windows

SpaceIm:
closes https://github.com/flexible-collision-library/fcl/issues/458

This big patch comes with several modifications:
- **FCL_EXPORT** is renamed **FCL_API**
- **FCL_EXPORT** (aka **FCL_API** now) is removed in all **template declarations**.
- 2 new definitions are introduced to handle explicit template declaration and explicit template definition for all compilers: **FCL_EXTERN_TEMPLATE_API** and **FCL_INSTANTIATION_DEF_API**. They are needed because Visual Studio expect dllexport in **explicit template definition**, while others compilers expect this kind of attribute in **explicit template declaration**.
These macros are defined through an option in generate_export_header (CUSTOM_CONTENT_FROM_VARIABLE) introduced in CMake 3.7. Therefore minimum version of CMake is bumped to 3.7 (maybe undesirable).
- It also adds more exported symbols.
- (my IDE has also removed a lot of trailing spaces).


SpaceIm:
___
*[CMakeLists.txt, line 34 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/461#-M4K-1hSBZYXh_IqHZCj:-M4VDPVuEJ_2DJN_kv-5:brjc8na) ([raw file](https://github.com/flexible-collision-library/fcl/blob/3b3a1245d9a4fd50f663e144a5909000db84a2c3/CMakeLists.txt#L34)):*
<details><summary><i>Previously, jamiesnape (Jamie Snape) wrote…</i></summary><blockquote>

If we want to continue to support Trusty, we need minimum 2.8.12, if we want to continue to support Xenial, we need minimum 3.5.

I am hoping to retool the build system for when we drop support for those platforms, but we are not at that stage yet, and I do not want to incrementally bump the CMake minimum version from 3.7 to 3.10 as we add features, since it adds complexity to our testing infrastructure (3.6-3.9 are not in an Ubuntu LTS).

Therefore, I am afraid, we cannot bump the version to 3.7.
</blockquote></details>

The only reason of 3.7 is the option CUSTOM_CONTENT_FROM_VARIABLE I use in generate_export_header in order to inject explicit template instantiation and definition definitions. There should be another way to achieve the same behaviour without this option.
For example, a less elegant solution could be to rename export.h, and create a file export.h with macros related to template instantiation which includes the file generated by generate_export_header. Thereby, we can rely of the current include behaviour of export.h in other files.

SpaceIm:
Honestly, I can't imagine redo all these modifications (357 files!) just to remove modifications related to trailing spaces :/

jwnimmer-tri:
Without implying any vote on whitespace process, FYI it's not difficult to fix.  First, start from master and remove all trailing whitespace in a new commit, PR that, and merge to master.  Then, rebase this PR atop that commit, always resolving conflicts by this PR winning (you can even just re-copy a backup of these files during the merge or rebase, don't even need to deal with diffs).

SpaceIm:
*[a discussion](https://reviewable.io/reviews/flexible-collision-library/fcl/461#-M4W21of3Zfzmg0VNRI9:-M4W6maF1xGAeJuT0YKU:b-vh89xu) (no related file):*

I didn't quite follow this:

>  Visual Studio expect dllexport in explicit template definition, while others compilers expect this kind of attribute in explicit template declaration.

Could you elaborate on what an "explicit template definition" is? What came to mind to me was [explicit instantiation](https://en.cppreference.com/w/cpp/language/class_template#Explicit_instantiation).

But then I'm not quite sure what "explicit template declaration" is.

</blockquote></details>

ooops sorry, it's a typo: I meant explicit instantiation declaration and explicit instantiation definition.

---

jwnimmer-tri:
It's simple to `git reset --hard OLDSHA` to reset this branch after a rebase.  It's O(1) not O(N).

jwnimmer-tri:
Rough outline:
Part A:
```
git checkout master
git checkout -b whitespace-only
perl -pi -e 's# *$##g;' $(find . -name '*.hh' -o -name '*.cc')
# git add, git commit
# open a new PR with whitespace-only branch; review, CI, and merge it to master
```

Part B:
```
git checkout fix-shared-windows  # this current PR
cp -r src ../src.backup
cp -r include ../include.backup
git checkout master
git checkout -b fix-shared-windows-without-whitespace
rm -rf src include
mv ../src.backup src
mv ../include.backup include
# git add, git commit
```

Now fix-shared-windows-without-whitespace is identical to this PR, except rebased on whitespace changes.

sherm1:
That seems very straightforward! An alternate part B could be
```
# After the part A whitespace is merged into master:
git checkout master
git pull upstream master
git checkout fix-shared-windows  # this current PR
git rebase master  # using your favorite merge tool
```
Merging should be easy.

ahoarau:
This PR works perfectly ! Thanks for the hard work. 
I would love to see that integrated in upstream.
Indeed `cmake` is staying with an old version on xenial, but having the latest 3.20.2 is very easy since it only requires to put the archive's bin in the `PATH`.

That being said, I also understand the need of having only `apt` executables. How long do you plan to "support" cmake 3.5 ?

jamiesnape:
> Indeed cmake is staying with an old version on `xenial`, but having the latest 3.20.2 is very easy since it only requires to put the archive's bin in the `PATH`.

Also you can use the `xenial` APT packages from https://apt.kitware.com.

> That being said, I also understand the need of having only `apt` executables. How long do you plan to "support" cmake 3.5 ?

I think we would be dropping support as soon as possible, but it has not been the highest priorities.

SeanCurtis-TRI (commented):
Do we have a straightforward solution to advancing this?

> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/461#-:-M_v25MH4paMfPkO_EN5:b-17064o)* status: all files reviewed, 3 unresolved discussions (waiting on @jamiesnape and @SpaceIm)

jamiesnape:
> Do we have a straightforward solution to advancing this?

At this stage, assuming that CI does not break: rebase, resolve conflicts, merge, and apologize to people using very old CMake (pre 3.7). If CI does break, I guess I could migrate it to GitHub Actions in an hour or two.

SteB0:
Hi, 
thanks god that i found this issue!
After struggling with fcl inside my windows project i found @SpaceIm´s branch. Is someone working on a solution to fix the master version?

SpaceIm:
Here is an attempt to solve merge conflict.

SpaceIm:
I've changed 2 more things:
- hide all symbols by default => there are no reason to not hide private symbols.
- allow to build tests when private symbols are hidden (I've also added `enable_testing()` at the root CMakeLists, it's convenient) => it works just fine.

sherm1:
I'd like to merge this heroic effort before code-rot sets in again! A few prerequisites
- FCL CI seems broken. Does anyone know how to fix it?
- We need at least one Windows user to verify that the most recent version solves the original problem. @SteB0 can you do that?

cc @SeanCurtis-TRI @jamiesnape @j-rivero @jslee02

sherm1:
Also cc @SeanCurtis-Drake

SteB0:
Hi, checked out latest version from @SpaceIm (265928f) and could build and use it without octomap and without Tests. After enabling "Build_Testing" in cmake-configuration i got linker errors e.g. 
`[46/81] Linking CXX executable test\test_fcl_collision.exe
FAILED: test/test_fcl_collision.exe`

`undefined reference to 'fcl::detail::ConvertBVImpl<double, fcl::AABB<double>, fcl::OBB<double> >::run(fcl::AABB<double> const&, Eigen::Transform<double, 3, 1, 0> const&, fcl::OBB<double>&)'`

SpaceIm:
Which compiler? I think it's fine with msvc, I'll try with a GNU compiler.

SteB0:
mingw810_64
x86_64-w64-mingw32-g++.exe

SpaceIm:
With apple-clang 13 on macOS, it works fine with `FCL_HIDE_ALL_SYMBOLS=ON`, `FCL_STATIC_LIBRARY=OFF` and `BUILD_TESTING=ON`.
I'll try with MinGW (but I have gcc 11.2.0 with runtime 9.0.0 on my windows computer, mingw810_64 is pretty old).

ahoarau:
Don't know if this is relevant for this MR, but I just found out that FCL can be made into a header-only library.
Can anyone explain the motivation behind declaring explicit templates in .cc ? Only speed when linking against it ?

SpaceIm:
I would say speed up compilation of downstream projects, by avoiding common template instantiations in fcl.

SpaceIm:
Ok so first issue to fix msvc shared with tests: turn `test_fcl_utility` to a static library.

SpaceIm:
Ok for Visual Studio, there was several missing default declaration & definition in `Contact` & `ContactPoint`. Moreover M_PI is not defined by default in cmath for Visual Studio, compilation of one test was failing.

SpaceIm:
For MinGW, there are indeed some link issues in tests. I'll see what I can do.

SpaceIm:
I think MinGW is fixed now with https://github.com/flexible-collision-library/fcl/pull/461/commits/a9ec3c2131208ab7fde1c602b0efb8c8dfc4c5e8

SpaceIm:
@SeanCurtis-TRI @jamiesnape @sherm1 I've fixed all the autoformatting mess, now only remain relevant modifications.

PR #463, opened 2020-04-15
Make computeAABB virtual in CollisionObject

Levi-Armstrong:
By making this virtual it allows for the developer to create a modified CollisionObject wrapper which adds a margin to the collision objects AABB.


Levi-Armstrong:
Another option would be to modify the Class to allow the user to set a margin on the AABB. Let me know if either of these are acceptable.

PR #467, opened 2020-04-29
Fix rss and improve octree tests

c-andy-martin:
There were several bugs in RSS handling that impacted octree distance checks.
There was also a bug in fcl::octree impacting when nodes were considered free/occupied.
Fix the bugs, and improve both RSS and octree distance check testing.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/467)

### DIFF
--- a/include/fcl/geometry/bvh/detail/BV_splitter-inl.h
+++ b/include/fcl/geometry/bvh/detail/BV_splitter-inl.h
@@ -531,34 +531,6 @@ struct ComputeSplitVectorImpl<S, kIOS<S>>
 {
   static void run(const kIOS<S>& bv, Vector3<S>& split_vector)
   {
-    /*
-      switch(bv.num_spheres)
-      {
-      case 1:
-      split_vector = Vector3<S>(1, 0, 0);
-      break;
-      case 3:
-      {
-      Vector3<S> v[3];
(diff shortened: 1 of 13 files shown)


c-andy-martin:
No worries. I've been away some for the 4th holiday. I think I have responded to everything.

c-andy-martin:
___
*[include/fcl/math/geometry.h, line 94 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/467#-MCRuftk9hKdDlHlc_4R:-MCRz2w75fMVN1wBNn1z:b-jt33vl) ([raw file](https://github.com/flexible-collision-library/fcl/blob/973388714c87b9a89e5c5220430383e6d288a1e5/include/fcl/math/geometry.h#L94)):*

BTW This has evolved so gradually to this point, that I've missed the elephant in the room. With the fact that you're passing in `x_axis` explicitly, there's no longer any need to do this C++03 style. We can bump it up to represent modern C++:

```c++
Matrix3<S> generateCoordinateSystem(const Vetctor3<S>& x_axis);
```

They've worked so hard to gives us the ever-so-convenient copy elision, let's make use of it.
</blockquote></details>

I agree, I will implement this today.

c-andy-martin:
OK, thanks @sherm1 for the review, I have addressed all your comments and everything should be taken care of.

PR #472, opened 2020-05-20
change contact, distance primitive id to `intptr_t`; add fcl::OcTree API to get info from opaque ids

c-andy-martin:
The distance result brief information stored the brief information
in an `int`. However, some primatives can only be stored by pointer
(namely octree nodes), and `int` is not wide enough on 64-bit
platforms to safely store a pointer. So upgrade the brief
information to be `intptr_t` which is guaranteed to be wide enough.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/472)

### DIFF
--- a/doc/Doxyfile.in
+++ b/doc/Doxyfile.in
@@ -2037,7 +2037,7 @@ INCLUDE_FILE_PATTERNS  =
 # recursively expanded use the := operator instead of the = operator.
 # This tag requires that the tag ENABLE_PREPROCESSING is set to YES.
 
-PREDEFINED             = FCL_DOXYGEN_CXX=1
+PREDEFINED             = FCL_DOXYGEN_CXX=1 FCL_HAVE_OCTOMAP
 
 # If the MACRO_EXPANSION and EXPAND_ONLY_PREDEF tags are set to YES then this
 # tag can be used to specify a list of macro names that should be expanded. The
--- a/include/fcl/common/types.h
+++ b/include/fcl/common/types.h
@@ -60,6 +60,8 @@ using int64 = std::int64_t;
 using uint64 = std::uint64_t;
 using int32 = std::int32_t;
 using uint32 = std::uint32_t;
+using intptr_t = std::intptr_t;
+using uintptr_t = std::uintptr_t;
 
 template <typename S>
 using Vector2 = Eigen::Matrix<S, 2, 1>;
--- a/include/fcl/geometry/octree/octree-inl.h
+++ b/include/fcl/geometry/octree/octree-inl.h
@@ -42,6 +42,8 @@
 
 #include "fcl/config.h"
 
+#include "fcl/geometry/shape/utility.h"
+
 #if FCL_HAVE_OCTOMAP
 
 namespace fcl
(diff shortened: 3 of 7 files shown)


c-andy-martin:
> Previously, SeanCurtis-TRI (Sean Curtis) wrote…
> At [this call site](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/narrowphase/detail/traversal/octree/octree_solver-inl.h#L297) It appears that the "id" of the node in question is the offset from the octree root to that node.
> 
> The same thing happens [here](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/narrowphase/detail/traversal/octree/octree_solver-inl.h#L509) and [here](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/narrowphase/detail/traversal/octree/octree_solver-inl.h#L787)
> 
> I presume from that that the tree is laid out in contiguous memory so the concept of a "node/cell id" seems meaningful. That makes it feel like passing pointers around is not the right thing.

No, octomaps are not laid out in contiguous memory. The nodes are allocated with `new`. Therefore, you can not take the id and turn it back into the node pointer using the root of the tree when it is cast to a 32-bit wide integer, as the upper bits of the address are lost and there is no guarantee that the node will be within a 32-bit distance of the tree's root node. I submitted this PR to fix that issue.

However, your comments bring up a good point that a valid node pointer does not directly give you the actual geometric box that was the closest, as the node pointer stores nothing more than the occupancy information of that cell, and the `fcl::Octree` does not expose any way of getting at the underlying octree (although the caller who created the octree could keep a pointer around to the queried tree). The only way you could figure out what box of space is being referred to from the id field would be to iterate over a bounding box of the tree, bounded by the returned distance, looking for the matching node pointer (inefficient).

I actually have a local change I run with which returns a shared pointer to the nearest primitives (and the corresponding transforms) but did not submit a PR for this because I only added support to return them for octomap/BVH distance queries. You can see that change [here](https://github.com/BadgerTechnologies/fcl/blob/fc4a905a241cb374bc117259e12f1fbe03e04296/include/fcl/narrowphase/distance_result.h#L89), and then I added returning that for octomap/BVH [here](https://github.com/BadgerTechnologies/fcl/blob/3bfcfddf476e428303f9ed39b3f3f30c2bced928/include/fcl/narrowphase/detail/traversal/octree/octree_solver-inl.h#L557). This way I can efficiently make use of the nearest box.

Yet, even though this change doesn't help much, it does correct the bug and does make it possible for the caller to find the correct octree cell, assuming they keep a pointer around to the tree they queried, and assuming they don't mind paying to iterate over the tree to find the correct cell. It does this at the cost of a small storage increase in the distance result and does also break the ABI for distance results, and I'm not sure what your policy is on that.

c-andy-martin:
OK. I'll code that up. I'm thinking tho instead of documenting the particulars of the id of the cell to provide an API in fcl::Octree that takes the ID and returns the cell information in various forms. I'll add some test cases to cover it.

c-andy-martin:
I have update w/ my implementation. Let me know if you'd like any changes, I'm especially wanting to be sure the public API names are acceptable.

c-andy-martin:
I'm not sure whether reviewable.io left my responses visible to you or not, but I have replied to your comments using reviewable.io and force-pushed a new version which combines all the API's into one getNode, hopefully addressing all your comments.

c-andy-martin:
___
*[include/fcl/geometry/octree/octree.h, line 155 at r3](https://reviewable.io/reviews/flexible-collision-library/fcl/472#-M8QaNEk1gHFmjSHBtcX:-M8Rgrua-7ENBozMDHsI:b-886gsx) ([raw file](https://github.com/flexible-collision-library/fcl/blob/ff3a0fab7862c02f76b9a2a6c529a50899bdba87/include/fcl/geometry/octree/octree.h#L155)):*

nit: These capitalization and final periods.
</blockquote></details>

Fixed

c-andy-martin:
___
*[include/fcl/geometry/octree/octree.h, line 161 at r3](https://reviewable.io/reviews/flexible-collision-library/fcl/472#-M8QaTuyAPiwL1HDe8cL:-M8RgumS6_b-5aGGPXQh:b355xk1) ([raw file](https://github.com/flexible-collision-library/fcl/blob/ff3a0fab7862c02f76b9a2a6c529a50899bdba87/include/fcl/geometry/octree/octree.h#L161)):*

nit: It now is apparent to me that this name should be more specific: `getNodeFromQueryid` (or some such thing).
</blockquote></details>

I chose `getNodeByQueryId` and updated accordingly.

c-andy-martin:
OK, I have addressed the comments. Also, in testing the Doxygen documentation to ensure my links from Contact and DistanceResult were working, I noticed that the OcTree documentation was not being generated at all. I have added another commit in the history which corrects this defect as well. (If you'd rather me open a separate PR for that I can).

c-andy-martin:
___
*[CMakeLists.txt, line 388 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/472#-M8SA51Q06SDMhuK5vW-:-M8TE-3s98XvwPj6jUfp:b-hfpv0l) ([raw file](https://github.com/flexible-collision-library/fcl/blob/71246691b3487cc8aaff43c3d09377e8df83e423/CMakeLists.txt#L388)):*

nit: This seems like a debugging artifact left over.
</blockquote></details>

removed

c-andy-martin:
___
*[CMakeLists.txt, line 386 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/472#-M8SLJQu3VMUzLo3cZTu:-M8TDyAV39n_E2VZGdWl:b-hfpv0l) ([raw file](https://github.com/flexible-collision-library/fcl/blob/71246691b3487cc8aaff43c3d09377e8df83e423/CMakeLists.txt#L386)):*

nit: I don't believe that this `env` prefix does anything. You've already configured the include path in the `Doxyfile.in`.  (Amusing that it wasn't previously used.)
</blockquote></details>

removed

c-andy-martin:
*[a discussion](https://reviewable.io/reviews/flexible-collision-library/fcl/472#-M8SBJhW6fxdK_e8pWwQ:-M8TBbYK5MXCE4CekMVz:b-rq6jcn) (no related file):*

I hadn't thought of the doxygen; that's a good call. It's spurred me to think about this and I'd like to have a quick back and forth to solicit your thoughts.

  - Are we doing the right thing?
    - What I see it doing is:
      - If FCL is successfully configured to use octomap, then building doxygen in that build environment will include things like `fcl::OcTree` and reference to that API found in `contact.h` and `distance_result.h` will create live links.
      - If FCL is not configured to use octomap (for any reason), the documentation in `contact.h` and `distance_result.h` will still reference the class and method, but no longer be a live link.
    - Are we happy with that intermittently dead link? Presumably, we'd expect the *de facto* documentation to be compiled with all the bells and whistles, so that those links would always be valid.
      - I can imagine if someone builds doxygen locally, that references should be valid regardless of how they configure it.
        - That suggests that we would handle these build-dependent connections differently.

I can imagine an alternative approach:
  - Doxygen should render all classes (regardless of what's included in the build).
  - Those classes/functions/whatevers that are only conditionally included in the binary should have some doxygen-renderable tag that communicates that they are conditional (with a link to instructions that would satisfy the condition).

If we were to do this, we'd simply define the missing macro `FCL_HAVE_OCTOMAP` in the `Doxyfile.in` file and include an alias in `Doxyfile.in` like `@octomap_conditional` that would get pushed into the doxygen of the appropriate entities which expands to the appropriate magical text.

This is rendered somewhat muddy because of how `FCL_HAVE_OCTOMAP` is used. In some cases (e.g. `octree.h`), it simply introduces/erases whole entities. In other files (e.g.,  `octree_distance_traversal_node.h`), the entities still exist, but including the file leads to a compile-time error. This requires some thought to decide if it increases the complexity. (And there are third cases where even *include* `octree.h` is guarded by a test on `FCL_HAVE_OCTOMAP` which is a little mysterious.)

I can imagine three possible outcomes here:

  1. Back the doxygen changes out of this PR and simply shrug.
  2. Include the doxygen changes included in this revision.
  3. Back the doxygen changes out and run up with a different solution.

Thoughts?
</blockquote></details>

I agree. So for FCL_HAVE_OCTOMAP, I think the best approach is to declare FCL_HAVE_OCTOMAP in Doxyfile.in. However, it didn't seem necessary for me to need to document the quirk that the OcTree class might not be delcared if octomap couldn't be found at compile time in more than one place, so I simply added a note to OcTree's documentation. I have rewritten the commit to be narrower and only deal w/ this issue for octomap (which seems relevant to address the missing documentation needed to document the new API method)

---

c-andy-martin:
___
*[include/fcl/geometry/octree/octree-inl.h, line 364 at r5](https://reviewable.io/reviews/flexible-collision-library/fcl/472#-M8X66E04BAzFsRkq9A_:-M8YDM3d81ApbxjTSA6u:bb6un1j) ([raw file](https://github.com/flexible-collision-library/fcl/blob/13fa2233db6bbf3e42af41ac0dfb16c3eb67c7d8/include/fcl/geometry/octree/octree-inl.h#L364)):*

Would be great to have that argument as a comment here in the code. Otherwise the next reader has to puzzle over it like I did to try to get convinced that it can't loop forever. It is a little jarring without some reassurance.
</blockquote></details>

Done

Issue #475, opened 2020-06-02
Incorrect declaration of virtual function makes project not compilable with float type


shumi12321:
https://github.com/flexible-collision-library/fcl/blob/ff8324929fdf76d6646d7b90c221a6a5bda438d8/include/fcl/math/motion/interp_motion.h#L79

This bug prevents building FCL types with float. Declaration of this function should not take double as argument, but templated type S.

shumi12321:
Also, here you are missing forward declaration for float:
https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/math/motion/interp_motion-inl.h#L48

sherm1:
Thanks @shumi12321 -- can you submit a PR to fix this?

shumi12321:
Hi @sherm1 , I creted following PR:
https://github.com/flexible-collision-library/fcl/pull/476

sherm1:
Thanks!

PR #476, opened 2020-06-03
Fixed issue with creating instances of type InterpMotion<float>

shumi12321:
In InterpMotion class,  method integrate() is not properly declared and defined which does not allow instancing of that class with type other than double.
This is proposed fix for this issue.

### DIFF
--- a/include/fcl/math/motion/interp_motion-inl.h
+++ b/include/fcl/math/motion/interp_motion-inl.h
@@ -45,6 +45,9 @@ namespace fcl
 
 //==============================================================================
 extern template
+class FCL_EXPORT InterpMotion<float>;
+
+extern template
 class FCL_EXPORT InterpMotion<double>;
 
 //==============================================================================
@@ -127,7 +130,7 @@ InterpMotion<S>::InterpMotion(
 
 //==============================================================================
 template <typename S>
-bool InterpMotion<S>::integrate(double dt) const
+bool InterpMotion<S>::integrate(S dt) const
 {
   if(dt > 1) dt = 1;
 
--- a/include/fcl/math/motion/interp_motion.h
+++ b/include/fcl/math/motion/interp_motion.h
@@ -76,7 +76,7 @@ public:
 
   /// @brief Integrate the motion from 0 to dt
   /// We compute the current transformation from zero point instead of from last integrate time, for precision.
-  bool integrate(double dt) const;
+  bool integrate(S dt) const;
 
   /// @brief Compute the motion bound for a bounding volume along a given direction n, which is defined in the visitor
   S computeMotionBound(const BVMotionBoundVisitor<S>& mb_visitor) const;


shumi12321:
Link to issue: https://github.com/flexible-collision-library/fcl/issues/475

shumi12321:
Also added same fix for ScrewMotion class.

shumi12321:
Added support for float type continuous collision objects.

shumi12321:
___
*[include/fcl/math/motion/interp_motion-inl.h, line 48 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8uVwy73WiEXXI0P3Wp:-M8ucZxG4XpMkeIDOrPL:bjgxhqi) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4df76bf198874f05b01ff46ed5da4704a6ae2796/include/fcl/math/motion/interp_motion-inl.h#L48)):*

nit: This is *not* a forward "declaration". This is informing the compiler that `InterpMotion<float>` has already been built and will be available to the linker from some arbitrary compilation object. However, this is not true. In fact, attempting to use an `InterpMotion<float>` will fail to link. This is not apparent because there are *zero* unit tests on this class.  :(

So, we have a number of things we can/should do:

1. Definitely remove this `extern template` declaration (and in all the other files where you've added it for `float`). This has to be done no matter what else we do.
2. This next is *optional* but highly preferred. We should add some unit tests for this class. That would make FCL distinctly better. (For example, I created the beginnings of that unit test and it revealed one bug already.) If you'd like the work I've done, then I can arrange to get you those changes (plenty of options there).
</blockquote></details>

I added definitions in .cpp files for those forward declarations. Is that ok or do you still want me to remove extern templates?

shumi12321:
*[a discussion](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8upt6r2jLaQreZ-RH0:-M8utaii6Mu-gIOyatB1:b3kgk00) (no related file):*

Regarding testing: I've got a commit that I can contribute to your PR.  It might save you some effort in setting up the boilerplate. Let me know if you want it and if you'd prefer receiving a patch, me pushing to your branch, whatever.
</blockquote></details>

That would be great :)

---

shumi12321:
*[a discussion](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8upt6r2jLaQreZ-RH0:-M8ynf3zDmClrVOrbOFN:b-ymtaz8) (no related file):*

Do you have a preference for how that contribution arises? I can:

- post a gist with the patch.
- submit a PR to your fork against this branch
- push directly to the branch (given permissions)
</blockquote></details>

I added you as collaborator on my forked repo so you can push it there to master branch.

---

shumi12321:
___
*[include/fcl/math/motion/interp_motion-inl.h, line 48 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8uVwy73WiEXXI0P3Wp:-M8ynoIZ5nzIxg9MMZek:bdbywmg) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4df76bf198874f05b01ff46ed5da4704a6ae2796/include/fcl/math/motion/interp_motion-inl.h#L48)):*

I believe the preference is *against* adding them. Here's the underlying belief and I'm open to being persuaded otherwise.

1. We assume that `double` is the most common scalar (so, we just include it in the library directly).
2. We assume that other scalars (e.g., `float`) are used sufficiently infrequently that we don't want to increase the library size and compile time for *everyone*.

So, the implication is that `float` users will be compiling their own instantiations from the template definition directly.

Thoughts?
</blockquote></details>

Ok, I will remove forward declaration for float types to speed up compiling for users who will not use floats.

shumi12321:
___
*[include/fcl/math/motion/screw_motion-inl.h, line 48 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8uX9WjEtm9CGHB93NK:-M8z8LpJ7Ch71lYzGOVj:ba3dfob) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4df76bf198874f05b01ff46ed5da4704a6ae2796/include/fcl/math/motion/screw_motion-inl.h#L48)):*

nit: Just marking the instances of `extern template` that need to be removed.
</blockquote></details>

Removed

shumi12321:
___
*[include/fcl/narrowphase/collision-inl.h, line 54 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8uXG1TArmBVKQY6EJQ:-M8z8PBQBQYmN7oc67Rb:ba3dfob) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4df76bf198874f05b01ff46ed5da4704a6ae2796/include/fcl/narrowphase/collision-inl.h#L54)):*

nit: Just marking the instances of `extern template` that need to be removed.
</blockquote></details>

Removed

shumi12321:
___
*[include/fcl/narrowphase/collision-inl.h, line 71 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8uXHD16xuJ-bV8fU-N:-M8z8SJK703ezKKxiFYn:ba3dfob) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4df76bf198874f05b01ff46ed5da4704a6ae2796/include/fcl/narrowphase/collision-inl.h#L71)):*

nit: Just marking the instances of `extern template` that need to be removed.
</blockquote></details>

Removed

shumi12321:
___
*[include/fcl/narrowphase/collision_object-inl.h, line 48 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8uXIW99J3f4W_wNwiV:-M8z8UYp-UHC6L5SWBIg:ba3dfob) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4df76bf198874f05b01ff46ed5da4704a6ae2796/include/fcl/narrowphase/collision_object-inl.h#L48)):*

nit: Just marking the instances of `extern template` that need to be removed.
</blockquote></details>

Removed

shumi12321:
___
*[include/fcl/narrowphase/continuous_collision-inl.h, line 59 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8uXL5o49JPgr-NN_Xz:-M8z8Wos8-Imu_aF4x9a:ba3dfob) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4df76bf198874f05b01ff46ed5da4704a6ae2796/include/fcl/narrowphase/continuous_collision-inl.h#L59)):*

nit: Just marking the instances of `extern template` that need to be removed.

(All of the changes to this file.)
</blockquote></details>

Removed

shumi12321:
___
*[src/math/motion/interp_motion.cpp, line 44 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8uXP9w0_Bua25eXITg:-M8z8Zmp8QDPeRQQ5TlM:b-frncuh) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4df76bf198874f05b01ff46ed5da4704a6ae2796/src/math/motion/interp_motion.cpp#L44)):*

nit: Just marking the instances of `extern template` that need to be removed.
</blockquote></details>

Not an extern template here.

shumi12321:
___
*[src/math/motion/screw_motion.cpp, line 44 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8uXQHPE1lyO1lFkBKP:-M8z8cpq9MC2eQiKzuXP:b-frncuh) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4df76bf198874f05b01ff46ed5da4704a6ae2796/src/math/motion/screw_motion.cpp#L44)):*

nit: Just marking the instances of `extern template` that need to be removed.
</blockquote></details>

Not an extern template here.

shumi12321:
___
*[src/narrowphase/collision.cpp, line 46 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8uXRuUE1LhYZjrV3tO:-M8z8e6q34KiiwM7RZFr:b-frncuh) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4df76bf198874f05b01ff46ed5da4704a6ae2796/src/narrowphase/collision.cpp#L46)):*

nit: Just marking the instances of `extern template` that need to be removed. (all of the changes of this file.)
</blockquote></details>

Not an extern template here.

shumi12321:
___
*[src/narrowphase/collision_object.cpp, line 44 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8uXVAl9x-ZpmfgfHMf:-M8z8fAi0AxRricrHZNN:b-frncuh) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4df76bf198874f05b01ff46ed5da4704a6ae2796/src/narrowphase/collision_object.cpp#L44)):*

nit: Just marking the instances of `extern template` that need to be removed.
</blockquote></details>

Not an extern template here.

shumi12321:
___
*[src/narrowphase/continuous_collision.cpp, line 45 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8uXWqn2Al2MdI69jA4:-M8z8gO55luyY1pxFn7t:b-frncuh) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4df76bf198874f05b01ff46ed5da4704a6ae2796/src/narrowphase/continuous_collision.cpp#L45)):*

nit: Just marking the instances of `extern template` that need to be removed.

All of the changes of this file.
</blockquote></details>

Not an extern template here.

shumi12321:
___
*[src/narrowphase/continuous_collision_object.cpp, line 44 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8uX_zY6hWnqRRMhGwu:-M8z8i152jm_9FdYYLGz:b-frncuh) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4df76bf198874f05b01ff46ed5da4704a6ae2796/src/narrowphase/continuous_collision_object.cpp#L44)):*

nit: Just marking the instances of `extern template` that need to be removed.
</blockquote></details>

Not an extern template here.

sherm1:
A random drive-by comment for @shumi12321:
People are usually disappointed when they switch from `double` to `float` hoping to improve performance. It is very common for code to run at the same speed or even slower in `float`. You give up a lot in accuracy with `float` computation, so you should definitely measure the results and not assume better performance. It is true that `float` will save memory over `double` for big meshes and sometimes that can also provide a performance improvement due to better cache locality. But an overall performance improvement is by no means guaranteed.

shumi12321:
*[a discussion](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8upt6r2jLaQreZ-RH0:-M9n0jWhE8pogoE3-BGd:bb70w76) (no related file):*

I've pushed a new commit:

1. It adds a skeleton unit test (specifically, I *started* testing constructors as a reality check).
    - Ideally, the constructor test would be finished for the five constructors.
    - Per *this* PR we definitely need to add a test showing that we can integrate `InterpMotion<S>` for `S = double, float`.
2. I also fixed some issues with the constructors that became apparent when I wrote the test -- it couldn't pass because the default constructor wasn't *really* initialized to identities. :-/
</blockquote></details>

Nice

---

shumi12321:
*[a discussion](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8znry9AA67uo0Vas_8:-M9n0onPBqLXWQTaz_zs:bv0qtjf) (no related file):*

I noticed you're new to github and I'm immensely pleased that we're your first contribution.  :)  As it appears you're new to this, I'd like to give you some lightweight guidance that'll make your life a bit easier.

1. You've made your modifications directly in your fork's `master` branch. Your life will be much simplified if you always do modifications in dedicated branches and PR from there. Right now, you're going to have some merging headaches in your *local* branch when we finally merge this PR to github's master.
    - The primary reason for this headache is we're going to squash all the commits down to a single commit. However, your master branch has a bunch of commits (which have the same *end* effect), but the fact that it's different commits will cause merge headaches. When we get to merging I can give you some guidance on how to update your master without too much headache.
    - Github has some nice documentation discussing this work flow: https://guides.github.com/introduction/flow/
2. We use reviewable to do code reviews -- it's got some of the nicest workflows for doing so. It comes along with some techniques and conventions that are *not* immediately clear.
    - Reviewable uses keywords like "nit", "minor", "btw", "fyi", "done". Starting a comment with one of these comments will have a special effect.
    - You'll notice that a lot of my comments start with "nit". That keyword indicates that I've flagged something minor with a simple resolution. I don't expect that it needs discussion. As such, if you resolve it, you can hit the "Done" button on the comment (or type "done") and the comment will automatically be marked as resolved -- no further work for me. :) The keyword "minor" has the same meaning as "nit".
    - If something is marked with "nit" or "minor" it is supposed to communicate that the reviewer is calling out a *defect* in your code. You may disagree it's a defect -- it happens all the time. At that point, responding with your reasoning is appropriate. Usually, one party persuades the other party and the resolution evolves organically.
    - The keywords "BTW" or "FYI" are intended as passing thoughts -- they should never be used to indicate a change you *should* make. As such, you are completely at liberty to simply disregard them (or, if it's a suggestion that you like, you are equally free to make a change). It's 100% up to you.
</blockquote></details>

1. The whole purpose of this fork is to submit this PR. I will remove it as soon as PR is done.
2. Thanks for this info, didn't know that

---

shumi12321:
___
*[include/fcl/math/motion/interp_motion.h, line 79 at r4](https://reviewable.io/reviews/flexible-collision-library/fcl/476#-M8uXjtI1oqD188Z9_Fe:-M9n1LnF-QLGaJAHhvIR:b-skq5mz) ([raw file](https://github.com/flexible-collision-library/fcl/blob/4df76bf198874f05b01ff46ed5da4704a6ae2796/include/fcl/math/motion/interp_motion.h#L79)):*

nit: This change is to enable something very specific: `InterpMotion<float>::integrate()`. No matter what else we do, we should introduce a unit test that proves that it works.
</blockquote></details>

Unit tests were added as another commit...

PR #488, opened 2020-08-04
Performance optimization for supportConvex

SeanCurtis-TRI:
The original implementation of `supportConvex` did a linear search across all vertices. This makes the Convex mesh smarter to make the search for the support vertex cheaper.

 - It builds an adjacency graph on all vertices upon construction.
 - It introduces the method `findExtremeVertex` method which does a walk along the edges of the mesh to find the extreme vertex.
   - This is faster for large meshes but slower for small meshes.
   - Empirical data suggests around 32 vertices to be a good cut off.
   - This uses vertex count to determine which algorithm to use.
 - It also introduces initial validation of the mesh -- that validation absolutely necessary for the edge walking to provide a correct answer.
 - Several aspects of the implementation have been tested and tweaked    for performance. See:
   - use of `vector<int8_t>` in `Convex::findExtremeVertex`
   - cache-friendly implementation of adjacency graph `Convex::neighbors_`.

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/488)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@
     [#467](https://github.com/flexible-collision-library/fcl/pull/467)
   * Bugs in RSS distance queries fixed:
     [#467](https://github.com/flexible-collision-library/fcl/pull/467)
+  * Convex gets *some* validation and improved support for the GJK `supportVertex()` API:
+    [#488](https://github.com/flexible-collision-library/fcl/pull/488)
 
 * Broadphase
 
--- a/include/fcl/geometry/shape/convex-inl.h
+++ b/include/fcl/geometry/shape/convex-inl.h
@@ -39,6 +39,10 @@
 #ifndef FCL_SHAPE_CONVEX_INL_H
 #define FCL_SHAPE_CONVEX_INL_H
 
+#include <map>
+#include <set>
+#include <utility>
+
 #include "fcl/geometry/shape/convex.h"
 
 namespace fcl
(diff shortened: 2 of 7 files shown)


jmirabel:
___
*[include/fcl/geometry/shape/convex-inl.h, line 268 at r3](https://reviewable.io/reviews/flexible-collision-library/fcl/488#-ME7CX7A5mu0wUaiOh21:-ME7CX7A5mu0wUaiOh22:b-snbccl) ([raw file](https://github.com/flexible-collision-library/fcl/blob/7b4285bab9d21d0b1d9b2e58759c1da2926b4727/include/fcl/geometry/shape/convex-inl.h#L268)):*
> ```objc
>   // Note: A small amount of empirical testing suggests that a vector of int8_t
>   // yields a slight performance improvement over int and bool. This is *not*
>   // definitive.
> ```

I suspect this is an arbitrage. `std::vector<bool>` is more cache-friendly than `int8_t` but there is an extra calculation which bit into which byte (i.e. `int8_t`) must be read. This depends on the number of vertices and on your hardware. Did you try with a large number of vertices (where large depends on your cache size) ?

jamiesnape:
Haven't looked into the error, the warnings are because you are explicitly choosing an ancient version of macOS that even Homebrew does not support.

Either:

* delete the line https://github.com/flexible-collision-library/fcl/blob/1bddc981de578d971cc59eb54f5d248c9d803b25/.travis.yml#L27 to get the default combination of macOS 10.13 and Xcode 9.4 (might still give a few warnings);
* change that line to `osx_image: xcode10.1` for macOS 10.13 and the last available Xcode for that OS version;
* change that line to `osx_image: xcode11.3` for macOS 10.14 and the last available Xcode for that OS version;
* change that line to `osx_image: xcode11.6` for macOS 10.15 and the latest stable Xcode.

Probably `osx_image: xcode11.3` is nearest to Drake, if that matters.

jamiesnape:
Since the OS is so old, Homebrew has to build everything from source and there was a new version of CMake in the last week that no longer compiles. At that point Travis CI looks to be falling back to an old CMake since there is no error checking in any of the CI scripts.

jamiesnape:
Note, if you choose Xcode 9.4 your build will break again in a month or so when Apple releases macOS 11 and support is dropped for macOS 10.13.

jamiesnape:
10.14 based on release date or the unreleased 11.0 based on end of support.

PR #514, opened 2020-12-08
Default enable SSE only if compiler supports it

### LINKED ISSUE #513: Build failure on ARM (ROS Noetic release issue)
copy-pasted from [here](https://discourse.ros.org/t/can-we-get-a-noetic-sync-please/17240/8)

@sloretz said:

> It looks like the [fcl build is still failing on arm](http://build.ros.org/view/Nbin_ufv8_uFv8/job/Nbin_ufv8_uFv8__fcl__ubuntu_focal_arm64__binary/), even [with the new patch checking if the host supports SSE](https://github.com/flexible-collision-library/fcl/commit/cbfe1e9405aa68138ed1a8f33736429b85500dea). The console output says it has the latest version ( `0.6.1-2focal.20201126.000025` ), so that patch wasn’t enough. It seems like cmake thinks the hosts supports SSE and SSE2, but the compiler does not. I think resolving the issue means another patch to fcl :frowning: followed by another release of `fcl` into Noetic, followed by a MoveIt release to use `fcl` again.

> For a patch to `fcl` I would guess using [check_cxx_compiler_flag()](https://cmake.org/cmake/help/latest/module/CheckCXXCompilerFlag.html) with the flags in [SSE_FLAGS](https://github.com/flexible-collision-library/fcl/blob/968fd7e0334a39e21e314959c60d5f21c13867ee/CMakeLists.txt#L98) to set the default value of the option [FCL_USE_x64_SSE](https://github.com/flexible-collision-library/fcl/blob/968fd7e0334a39e21e314959c60d5f21c13867ee/CMakeLists.txt#L95) could resolve the issue.

sloretz:
This fixes an issue seen in ROS Noetic where SSE is default enabled on
an aarch64 system. CMake seems to think the host supports SSE, but the
compiler does not.

I don't know the root cause of the issue. Maybe it's because the ROS 1 buildfarm uses qemu on x86 systems, and the method CMake uses to determine SSE support shows that the machine beneath qemu supports it?

Fixes #513

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/514)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,30 +79,40 @@ else()
   option(FCL_STATIC_LIBRARY "Whether the FCL library should be static rather than shared" OFF)
 endif()
 
+set(SSE_FLAGS "")
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  set(SSE_FLAGS -mfpmath=sse -msse -msse2 -msse3 -mssse3)
+elseif(MSVC)
+  # Win64 will add the flag automatically
+  if(CMAKE_VS_PLATFORM_NAME STREQUAL "Win32")
+    set(SSE_FLAGS /arch:SSE2)
+  endif()
+endif()
+
 # Whether to enable SSE
 if(CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
   set(FCL_TARGET_SUPPORT_X64_SSE OFF)
 else()
+  # Does the host support SSE
   cmake_host_system_information(RESULT _has_sse QUERY HAS_SSE)
   cmake_host_system_information(RESULT _has_sse2 QUERY HAS_SSE2)
-  if(_has_sse AND _has_sse2)
+
+  # Does the compiler support SSE
+  include(CheckCXXCompilerFlag)
+  check_cxx_compiler_flag("${SSE_FLAGS}" _compiler_supports_sse)
+
+  if(_has_sse AND _has_sse2 AND _compiler_supports_sse)
     set(FCL_TARGET_SUPPORT_X64_SSE ON)
   else()
     set(FCL_TARGET_SUPPORT_X64_SSE OFF)
   endif()
 endif()
-set(SSE_FLAGS "")
 option(FCL_USE_X64_SSE "Whether FCL should x64 SSE instructions" ${FCL_TARGET_SUPPORT_X64_SSE})
 if(FCL_USE_X64_SSE)
-  if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-    set(SSE_FLAGS -mfpmath=sse -msse -msse2 -msse3 -mssse3)
-  elseif(MSVC)
-    # Win64 will add the flag automatically
-    if(CMAKE_VS_PLATFORM_NAME STREQUAL "Win32")
-      set(SSE_FLAGS /arch:SSE2)
-    endif()
-  endif()
   add_compile_options(${SSE_FLAGS})
+else()
+  # Unset SSE_FLAGS so it doesn't get used in subdirectories
+  set(SSE_FLAGS "")
 endif()
 
 option(FCL_USE_HOST_NATIVE_ARCH "Whether FCL should use cflags from the host used to compile" OFF)


SeanCurtis-TRI (approved):
:LGTM:
+@jamiesnape Could you take a once over on this one? It makes sense to me: ultiamtely, it distinguishes between what the *host* reports vs the *compiler*.

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/514#-:-MO3lCtf25GYRQS0yo4H:b7mpif2)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jamiesnape)

sherm1 (approved):
Also :lgtm: with a question below. Would be great to get a quick look from @jamiesnape -- just a small change here.

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/514#-:-MO7TMwr3oIxZiuveBZi:b-ifi398)* status: all files reviewed, 1 unresolved discussion (waiting on @jamiesnape and @sloretz)

___
*[CMakeLists.txt, line 115 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/514#-MO7SegP2OjvAOOqJuL1:-MO7SegP2OjvAOOqJuL2:b-ggkv7m) ([raw file](https://github.com/flexible-collision-library/fcl/blob/81d797677e2e9fc1918dac7cd71cac233644cc3f/CMakeLists.txt#L115)):*
> ```CMake
> else()
>   # Unset SSE_FLAGS so it doesn't get used in subdirectories
>   set(SSE_FLAGS "")
> ```

Just curious -- @sloretz why unset this particular variable? Does it have some global meaning that other local variables like FCL_TARGET_SUPPORT_X64_SSE do not?

SeanCurtis-TRI (approved):
> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/514)* status: all files reviewed, 1 unresolved discussion (waiting on @jamiesnape and @sloretz)

___
*[CMakeLists.txt, line 115 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/514#-MO7SegP2OjvAOOqJuL1:-MO7VLdE6VrQ_ry731bl:bdv8pfa) ([raw file](https://github.com/flexible-collision-library/fcl/blob/81d797677e2e9fc1918dac7cd71cac233644cc3f/CMakeLists.txt#L115)):*

Just curious -- @sloretz why unset this particular variable? Does it have some global meaning that other local variables like FCL_TARGET_SUPPORT_X64_SSE do not?
</blockquote></details>

That's funny. I had the same question, and then I realized he *sets* those flags based on host to perform a compiler test. The result of that test is stored in the `_compiler_supports_sse` variable. If we determine that the compiler doesn't support sse, we then need to clear that variable so it doesn't pollute, as the comment says, subdirectories `CMakeLists.txt` contexts.

I spent a brief moment considering if a different articulation of the logic would eliminate the need to clear these flags again, but decided there probably isn't (or, at least, not obviously). It's rendered somewhat confusing because there are a numb of SSE related symbols all churning around.

SeanCurtis-TRI (approved):
> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/514)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jamiesnape)

SeanCurtis-TRI (approved):
> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/514)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jamiesnape)

jamiesnape (approved):
This code is getting really ugly (not the fault of he author of this PR) and I wonder whether a library should ever be setting SSE flags, but aside from my comment the code is an improvement, so :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/514#-:-MO7VqJh2HaIKjHO99CM:b-2g5em0)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sloretz)

___
*[CMakeLists.txt, line 115 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/514#-MO7SegP2OjvAOOqJuL1:-MO7UAtb37tIlU_85kun:b-gfukao) ([raw file](https://github.com/flexible-collision-library/fcl/blob/81d797677e2e9fc1918dac7cd71cac233644cc3f/CMakeLists.txt#L115)):*

Just curious -- @sloretz why unset this particular variable? Does it have some global meaning that other local variables like FCL_TARGET_SUPPORT_X64_SSE do not?
</blockquote></details>

Looks like it is used in `src/CMakeLists.txt`. Not sure why we are setting it globally and per-target, however. I think I would wrap the usage there with `if(FCL_USE_X64_SSE) ... endif()` if it is not already set.

sherm1 (approved):
> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/514)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sloretz)

___
*[CMakeLists.txt, line 115 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/514#-MO7SegP2OjvAOOqJuL1:-MO7chZ_9oYFMQETgq85:bqjqw4s) ([raw file](https://github.com/flexible-collision-library/fcl/blob/81d797677e2e9fc1918dac7cd71cac233644cc3f/CMakeLists.txt#L115)):*
<details><summary><i>Previously, jamiesnape (Jamie Snape) wrote…</i></summary><blockquote>

Looks like it is used in `src/CMakeLists.txt`. Not sure why we are setting it globally and per-target, however. I think I would wrap the usage there with `if(FCL_USE_X64_SSE) ... endif()` if it is not already set.
</blockquote></details>

Oh, I see. A cleaner alternative might be not to use SSE_FLAGS here to hold the proposed flags -- how about using a local variable like `_proposed_sse_flags` and then set SSE_FLAGS only if the host and compiler tests both pass?

sloretz:
> Just curious -- @sloretz why unset this particular variable? Does it have some global meaning that other local variables like FCL_TARGET_SUPPORT_X64_SSE do not?

No special meaning other than it's used in `src/CMakeLists.txt`

> Looks like it is used in src/CMakeLists.txt. Not sure why we are setting it globally and per-target, however. I think I would wrap the usage there with if(FCL_USE_X64_SSE) ... endif() if it is not already set.

Yup, link below:
https://github.com/flexible-collision-library/fcl/blob/968fd7e0334a39e21e314959c60d5f21c13867ee/src/CMakeLists.txt#L51

> Oh, I see. A cleaner alternative might be not to use SSE_FLAGS here to hold the proposed flags -- how about using a local variable like _proposed_sse_flags and then set SSE_FLAGS only if the host and compiler tests both pass?

I'm happy to make any changes to the logic there's consensus on.

sherm1 (approved):
Thoughts @jamiesnape & @seancurtis-tri ?

> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/514#-:-MO8_T2a0UQv4C25sG-v:bsu912g)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sloretz)

SeanCurtis-TRI (approved):
I do like the idea of setting a local variable with the flags and only setting `SSE_FLAGS` if we *mean* to do so. The "set-and-unset-as-necessary" paradigm would be a bit brittle. That said, is it worth it? After all, it's one small rock on top of a pile of ugly rocks. The real thing would be to have someone at @jamiesnape's level actually go through and make this sane.

Conclusion: I'm content to take it as is (having passed CI) and defer polishing the rock to later.

> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/514#-:-MO93WTGEDlTsRGsKzH7:b-kjtyfe)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sloretz)

jamiesnape (approved):
Let's merge then. This is not the biggest problem with the build system.

I probably will have some time at some point to redesign the build for modern CMake. Really the build on this project should be clean and simple (one library, some tests, and relatively few dependencies); it has just grown organically over a very long time trying to support old versions of CMake and many versions of third--party dependencies. That support is probably no longer needed, which will make cleaning up fairly quick.

> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/514#-:-MOHpEp8ALzToBPHV-XS:b-zdpts1)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sloretz)

SeanCurtis-TRI (approved):
Sounds like something to look forward to.

> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/514#-:-MOHvgq94AraZmMRuNvV:baoygtg)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sloretz)

rhaschke:
@sloretz, I'm going to prepare a new ROS release.

PR #550, opened 2021-10-16
fix collision between mesh mesh BVH nodes

weshoke:
Two boxes translated towards each other fail `continuousCollide`.   When the bounds in `MeshConservativeAdvancementTraversalNode<BV>::leafTesting is computed`, the bounds values are signed.  When summed together, they can offset leaving the value for `cur_delta_t` incorrect.

If their absolute values are summed, the test case below succeeds.

```
#include "fcl/math/bv/utility.h"
#include "fcl/narrowphase/collision.h"
#include "fcl/narrowphase/continuous_collision.h"
#include "fcl/narrowphase/detail/gjk_solver_indep.h"
#include "fcl/narrowphase/detail/gjk_solver_libccd.h"
#include "fcl/narrowphase/detail/traversal/collision_node.h"
#include "fcl_resources/config.h"
#include "test_fcl_utility.h"

using BVHModel = fcl::BVHModel<fcl::OBBRSSd>;
using Vector3 = fcl::Vector3<double>;
using Transform3 = fcl::Transform3<double>;
using CollisionObject = fcl::CollisionObject<double>;
using CollisionRequest = fcl::CollisionRequest<double>;
using CollisionResult = fcl::CollisionResult<double>;
using ContinuousCollisionRequest = fcl::ContinuousCollisionRequest<double>;
using ContinuousCollisionResult = fcl::ContinuousCollisionResult<double>;

int main(int argc, char* argv[]) {
  auto verts1 =
      std::vector<Vector3>{Vector3(-1.0, -1.0, 3.0), Vector3(1.0, -1.0, 3.0),
                           Vector3(-1.0, 1.0, 3.0),  Vector3(1.0, 1.0, 3.0),
                           Vector3(-1.0, -1.0, 4.0), Vector3(1.0, -1.0, 4.0),
                           Vector3(-1.0, 1.0, 4.0),  Vector3(1.0, 1.0, 4.0)};
  auto verts2 = std::vector<Vector3>{
      Vector3(-10.0, -10.0, 0.0), Vector3(10.0, -10.0, 0.0),
      Vector3(-10.0, 10.0, 0.0),  Vector3(10.0, 10.0, 0.0),
      Vector3(-10.0, -10.0, 0.5), Vector3(10.0, -10.0, 0.5),
      Vector3(-10.0, 10.0, 0.5),  Vector3(10.0, 10.0, 0.5)};

  auto faces = std::vector<fcl::Triangle>{
      fcl::Triangle(3, 1, 0), fcl::Triangle(3, 0, 2), fcl::Triangle(0, 1, 5),
      fcl::Triangle(0, 5, 4), fcl::Triangle(1, 3, 7), fcl::Triangle(1, 7, 5),
      fcl::Triangle(3, 2, 6), fcl::Triangle(3, 6, 7), fcl::Triangle(2, 0, 4),
      fcl::Triangle(2, 4, 6), fcl::Triangle(4, 5, 7), fcl::Triangle(4, 7, 6),
  };

  auto bvh1 = std::make_shared<BVHModel>();
  bvh1->beginModel(8, 12);
  bvh1->addSubModel(verts1, faces);
  bvh1->endModel();

  auto bvh2 = std::make_shared<BVHModel>();
  bvh2->beginModel(8, 12);
  bvh2->addSubModel(verts2, faces);
  bvh2->endModel();

  auto tf11 = Transform3();
  tf11.setIdentity();
  tf11.translation()[0] = 1.0;
  tf11.translation()[1] = 0.0;
  tf11.translation()[2] = -1.0;

  auto tf12 = Transform3();
  tf12.setIdentity();
  tf12.translation()[0] = 1.0;
  tf12.translation()[1] = 0.0;
  tf12.translation()[2] = -2.0;

  auto tf21 = Transform3();
  tf21.setIdentity();
  tf21.translation()[0] = -2.0;
  tf21.translation()[1] = 0.0;
  tf21.translation()[2] = 0.0;

  auto tf22 = Transform3();
  tf22.setIdentity();
  tf22.translation()[0] = 2.0;
  tf22.translation()[1] = 0.0;
  tf22.translation()[2] = 1.5;

  const auto* obj1 = new CollisionObject(bvh1, tf11);
  const auto* obj2 = new CollisionObject(bvh2, tf21);

  auto request = ContinuousCollisionRequest(
      100, 1e-4, fcl::CCDMotionType::CCDM_TRANS, fcl::GJKSolverType::GST_INDEP,
      fcl::CCDSolverType::CCDC_CONSERVATIVE_ADVANCEMENT);

  auto result = ContinuousCollisionResult();
  auto ret = fcl::continuousCollide(obj1, tf12, obj2, tf22, request, result);
  std::cout << "is_collide: " << result.is_collide << "\n";
  std::cout << "time_of_contact: " << result.time_of_contact << "\n";

  return 0;
}
```

Master output:

```
is_collide: 0
time_of_contact: 1 
```

This branch's output:

```
is_collide: 1
time_of_contact: 0.6
```

---
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/550)

### DIFF
--- a/include/fcl/narrowphase/detail/traversal/distance/mesh_conservative_advancement_traversal_node-inl.h
+++ b/include/fcl/narrowphase/detail/traversal/distance/mesh_conservative_advancement_traversal_node-inl.h
@@ -38,41 +38,51 @@
 #ifndef FCL_TRAVERSAL_MESHCONSERVATIVEADVANCEMENTTRAVERSALNODE_INL_H
 #define FCL_TRAVERSAL_MESHCONSERVATIVEADVANCEMENTTRAVERSALNODE_INL_H
 
-#include "fcl/math/bv/RSS.h"
-#include "fcl/math/motion/triangle_motion_bound_visitor.h"
 #include "fcl/narrowphase/detail/traversal/distance/mesh_conservative_advancement_traversal_node.h"
 
-namespace fcl {
+#include "fcl/math/bv/RSS.h"
+#include "fcl/math/motion/triangle_motion_bound_visitor.h"
 
-namespace detail {
+namespace fcl
+{
+
+namespace detail
+{
 
 //==============================================================================
-extern template class FCL_EXPORT
-    MeshConservativeAdvancementTraversalNodeRSS<double>;
+extern template
+class FCL_EXPORT MeshConservativeAdvancementTraversalNodeRSS<double>;
 
 //==============================================================================
-extern template bool initialize(
+extern template
+bool initialize(
     MeshConservativeAdvancementTraversalNodeRSS<double>& node,
-    const BVHModel<RSS<double>>& model1, const Transform3<double>& tf1,
-    const BVHModel<RSS<double>>& model2, const Transform3<double>& tf2,
+    const BVHModel<RSS<double>>& model1,
+    const Transform3<double>& tf1,
+    const BVHModel<RSS<double>>& model2,
+    const Transform3<double>& tf2,
     double w);
 
 //==============================================================================
-extern template class FCL_EXPORT
-    MeshConservativeAdvancementTraversalNodeOBBRSS<double>;
+extern template
+class FCL_EXPORT MeshConservativeAdvancementTraversalNodeOBBRSS<double>;
 
 //==============================================================================
-extern template bool initialize(
+extern template
+bool initialize(
     MeshConservativeAdvancementTraversalNodeOBBRSS<double>& node,
-    const BVHModel<OBBRSS<double>>& model1, const Transform3<double>& tf1,
-    const BVHModel<OBBRSS<double>>& model2, const Transform3<double>& tf2,
+    const BVHModel<OBBRSS<double>>& model1,
+    const Transform3<double>& tf1,
+    const BVHModel<OBBRSS<double>>& model2,
+    const Transform3<double>& tf2,
     double w);
 
 //==============================================================================
 template <typename BV>
-MeshConservativeAdvancementTraversalNode<
-    BV>::MeshConservativeAdvancementTraversalNode(typename BV::S w_)
-    : MeshDistanceTraversalNode<BV>() {
+MeshConservativeAdvancementTraversalNode<BV>::
+MeshConservativeAdvancementTraversalNode(typename BV::S w_)
+  : MeshDistanceTraversalNode<BV>()
+{
   delta_t = 1;
   toc = 0;
   t_err = (S)0.00001;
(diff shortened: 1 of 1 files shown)


sherm1 (commented):
I'd like to get this one-line fix merged if it's right! But it's not clear to me what the meaning of a negative bound is here. Can you explain why taking the absolute value is the right thing to do?

@seancurtis-tri what do you think?

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/550#-:-MphFFN-05XvTViTfEdK:b-qsayxz)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @sherm1)

SeanCurtis-TRI (commented):
I agree, it's very difficult to reason about. It's not clear what the "motion bound" quantity actually refers to. I've done a bit of spelunking, and this quantity depends on the type of object being "bounded" (in the code below, it's triangle, based on the use of `TriangleMotionBoundVisitor`). It also depends on the motion implementation (screw, translation, interpolation, etc.)

If you [look at the simplest version (translational motion)](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/math/motion/triangle_motion_bound_visitor-inl.h#L225), the bound is simply the "motion's velocity" dotted with the visitor's "normal". 

In the code below, two visitors are created: one one the first triangle's vertices and one on the second triangle's vertices. They are provided the *exact same normal*. That means, if the two triangles are moving in opposite directions at the same speed, they'll produce "motion bound" values that are exactly opposites of each other and the sum goes to zero. Conversely, if their relative velocity were zero, we'd get a motion bound value proportionate to their common speed. Even without knowing with the "motion bound" is, that seems surprising at best. It makes *me* wonder if either a) the normal shouldn't be reversed for the second visitor, or b) the bound should be their difference (just as the `n = P2 - P1`, `bound = bound2 - bound1` (or some such thing).

Generally, the absolute values seem highly suspicious.

> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/550#-:-MphMWN-4iYJ9WwMBY1J:bgx46pm)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @sherm1)

sherm1 (commented):
Agreed. @weshoke it would be great to get this fixed but would take more analysis to determine and document the correct behavior, and a few unit tests to verify it. Are you up for doing that work? If not it would be great to have an issue filed that clearly documents the problem so that someone can tackle it later -- the issue itself would be a useful contribution.

> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/550#-:-MphRhn0Bwv0NLWvfyEU:b-attgba)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @sherm1)

weshoke:
@sherm1 I'm happy to make an issue.  I don't have the cycles at this point to go further and try to address it properly.  I was testing out FCL to see if it could be used in a 5-Axis CNC simulation.  Unfortunately I found too many issues where it was missing collisions that I moved on to other solutions.  In general, I found the mesh collision routines to be quite buggy.  This was the simplest case, but when looking at screw motions, I found similar results.  That code was too complicated for me to dig into though.

Please don't take the above as me knocking FCL.  I like the project, but it's not something I found was robust enough for my use case.  It seemed like non-mesh collisions were more correct and not as much effort had gone in to testing the mesh capabilities.

weshoke:
Here's the above as an issue https://github.com/flexible-collision-library/fcl/issues/555

sherm1:
Great, thanks Wesley. Knocking FCL is fine with me -- we do it all the time! It has had many robustness issues. We've been repairing the ones we need for our own work but that is only a subset of FCL functionality. It's not surprising to hear there are problems in areas we haven't touched.

PR #566, opened 2022-01-30
Fix circular inclusion of triangle_motion_bound_visitor.h

SpaceIm:
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/566)


SpaceIm:
Actually these headers are required, it breaks compilation. But circular includes should be removed in a way or another.

PR #569, opened 2022-02-28
Fix errors in sphereSphere and sphereTriangle intersect calculation

shshlei:
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/569)


shshlei:
For intersect detection, the penetration depth is calculated as a positive value.

SeanCurtis-TRI (changes requested):
Thanks for contributing. Some further clarity on this PR would be very helpful. Of the three changes here:

1. sphere-triangle look straightforward and clear. The penetration depth value is clearly coming out negative when it shouldn't.
2. For sphere-sphere
    - the "contact point" calculation seems a definite improvement. The fact that the old code exists at all (and hasn't given rise to bugs already) shocks me. I think we're missing unit tests.
    - The change of contact normal for when sphere are coincident needs some further justification. See my notes below.

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/569#-:-Mwyf9EC6bM-l53sjgpp:bqvv0bn)* status: all files reviewed, 6 unresolved discussions (waiting on @shshlei)

*[a discussion](https://reviewable.io/reviews/flexible-collision-library/fcl/569#-Mwye5NG38UzMYCsweW5:-Mwye5NG38UzMYCsweW6:bcp5ncu) (no related file):*
You should add these fixes to the [change log](https://github.com/flexible-collision-library/fcl/blob/master/CHANGELOG.md).

---
___
*[-- commits, line 2 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/569#-MwyeEzZ25wz0AEwdPKk):*
Based on my concerns on clarity about what this PR actually does, it would be helpful to everyone to capture that in the commit message.
___
*[include/fcl/narrowphase/detail/primitive_shape_algorithm/sphere_sphere-inl.h, line 79 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/569#-Mwyc2wQBxIEviPCUB41:-Mwyc2wQBxIEviPCUB42:bixelfa) ([raw file](https://github.com/flexible-collision-library/fcl/blob/161a9078eb9932ed43d021635ad1493bd46f6ef5/include/fcl/narrowphase/detail/primitive_shape_algorithm/sphere_sphere-inl.h#L79)):*
> ```objc
>     // If the centers of two sphere are at the same position, the normal is (1, 0, 0).
>     // Otherwise, normal is pointing from center of object 1 to center of object 2
>     const Vector3<S> normal = len > 0 ? (diff / len).eval() : Vector3<S>(1, 0, 0);
> ```

I don't know that you get to unilaterally change the behavior of the normal for coincident spheres. The documentation says it returns the zero vector, and that's what the code was returning.

Changing the behavior of the function now can break other code that may explicitly or implicitly rely on this behavior.

I can certainly understand not liking a contact normal that *may* be zero. But changing it here should only be done if you're convinced that it's not zero anywhere else. Does this change make FCL more consistent or less consistent? (Sorry, I don't have the cycles to answer that question myself right now.)
___
*[include/fcl/narrowphase/detail/primitive_shape_algorithm/sphere_sphere-inl.h, line 80 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/569#-Mwyd5Nf3OWqxsNRQxmU:-Mwyd5Nf3OWqxsNRQxmV:b2u77hf) ([raw file](https://github.com/flexible-collision-library/fcl/blob/161a9078eb9932ed43d021635ad1493bd46f6ef5/include/fcl/narrowphase/detail/primitive_shape_algorithm/sphere_sphere-inl.h#L80)):*
> ```objc
>     // Otherwise, normal is pointing from center of object 1 to center of object 2
>     const Vector3<S> normal = len > 0 ? (diff / len).eval() : Vector3<S>(1, 0, 0);
>     const Vector3<S> point = tf2.translation() - s2.radius * normal;
> ```

This likewise defines a very different point from what was previously defined. This is a point *definitely* on the surface of sphere 2.

The old point is...more difficult to reason about. One would hope that it's just a really *weird* equivalent expression. But I'm confident it isn't. The old point doesn't lie on the surface of either sphere (generally). It *definitely* lies inside sphere 1. We know that `|diff| < r1 + r2` (that's the definition of being in collision), therefore we're always displacing from sphere 1's origin less than sphere 1's radius. And, while it's that a point on sphere 2's surface would be inside sphere 1's volume, I'm quite convinced that the expression being replaced isn't the right point.

For such a radical change, I'd expect to see unit tests fail (supposing they're defined). If not, we should have some unit tests added as part of this PR.

So, this seems like a definite improvement. I would ask the question about consistency. Off the top of my mind, I don't feel confident in asserting whether these primitive tests *generally* define the point on the 2nd object or the 1st. If the second, we're done. Great! Can you confirm?

*** as a foot note ***

I have to admit I'm kind of shocked to see this old line of code. I would've thought that the point not lying on the surface of one of the sphere's would've leapt out at *someone*. I certainly would've thought this would be trivially under unit test. My mind is blown that it isn't.
___
*[include/fcl/narrowphase/detail/primitive_shape_algorithm/sphere_triangle-inl.h, line 197 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/569#-MwyiKAPBYFTo0WY8icy:-MwyiKAPBYFTo0WY8icz:b-fvhxld) ([raw file](https://github.com/flexible-collision-library/fcl/blob/161a9078eb9932ed43d021635ad1493bd46f6ef5/include/fcl/narrowphase/detail/primitive_shape_algorithm/sphere_triangle-inl.h#L197)):*
> ```objc
>   if(has_contact)
>   {
>     Vector3<S> contact_to_center = contact_point - center;
> ```

BTW This is a *horrible* name for this variable. It's clearly the displacement vector *from* the center *to* the contact point.  :(

You're not obliged to fix this, but as long as you're here, it might be nice.  :)
___
*[include/fcl/narrowphase/detail/primitive_shape_algorithm/sphere_triangle-inl.h, line 205 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/569#-MwyijCX73w5lBJstbgg:-MwyijCX73w5lBJstbgh:bwc7fjp) ([raw file](https://github.com/flexible-collision-library/fcl/blob/161a9078eb9932ed43d021635ad1493bd46f6ef5/include/fcl/narrowphase/detail/primitive_shape_algorithm/sphere_triangle-inl.h#L205)):*
> ```objc
>       {
>         S distance = std::sqrt(distance_sqr);
>         if(normal_) *normal_ = contact_to_center.normalized();
> ```

BTW Given we just paid the cost of computing `distance`, this would be much better as:

```c++
if (normal_) *normal_ = contact_to_center / distance;
```

You are likewise not obliged to change this, but as long as you're here.

SeanCurtis-TRI:
@shshlei 

FTR This PR seems like it's no longer the original PR. It appears you've been pushing commits to your master and every time you push it back to your remote repo, this PR is growing in scope and scale.

The original comments were never resolved. Unless I hear from the author in the next few days, I'm going to go ahead and close the PR.

SeanCurtis-TRI:
The original PR was never finalized and now it has become polluted with subsequent, unrelated changes. If you want to pursue the original intent, please open another PR.

PR #614, opened 2024-04-11
Add support for Halfspace distance function call

KolinGuo:
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/614)


KolinGuo:
Open by mistake...

PR #626, opened 2024-10-11
[ci] Update macos images to the latest

jslee02:
`macos-11` was removed a long time ago, and macos-14 (ARM) and macos-14-large (x86) are the latest images:

https://github.com/actions/runner-images?tab=readme-ov-file#available-images

### DIFF
--- a/.github/workflows/all_ci.yml
+++ b/.github/workflows/all_ci.yml
@@ -44,7 +44,7 @@ jobs:
       matrix:
         compiler: [clang]
         build: [Release, Debug]
-        os: ['macos-11']
+        os: ['macos-14', 'macos-14-large']
     uses: ./.github/workflows/generic_ci.yml
     with:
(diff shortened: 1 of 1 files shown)


SeanCurtis-TRI:
Presumably this failed because the "large" runners aren't available for generic open source projects. Sadly, the CI logs are gone. But what remains includes the message:

> The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings.

I.e., you can't have large, you haven't paid. I would hope that the normal-sized machine would work. Could you try pushing again for `macos-14` and I can see if 14 is having the same linking problems I'm experiencing in `macos-latest` .

SeanCurtis-TRI:
Nevermind; I tried `macos-14` and the link problems I'm having are present there as well.

I propose we close this in favor of #635 because that also includes some changes to code to make newer compilers happy.

PR #658, opened 2026-01-22
Add real world convex-convex regression test

SeanCurtis-TRI:
The focus of this fix is how EPA computes the outward facing normal for a polytope face: faceNormalPointingOutward().

1. This function's ability to *meaningfully* pick an outward normal depends on how easy it is to classify the side of a face on which polytope vertices lie. This classification works best when we have some real *volume* in the polytope. - When creating a polytope from a *triangle*, rather than picking the fourth vertex based on its distance from the *triangle*, we pick based on the distance to the triangle plane. - By picking distance to triangle, we may simply pick a vertex that is co-planar with the triangle, but far removed (which renders outward normal computation random). - update convert2SimplexToTetrahedron() to reflect this.
2. faceNormalPointingOutward() was unconditionally normalizing the face normal but return the unnormalized vector.
   - Most call sites then had to subsequently normalize it.
   - We'll simply return the unit normal and not do redundant work.
   - This is incidental to the bug, but obvious and worthwhile.
3. faceNormalPointingOutward() has a test where the tet is sufficiently flat where we compare max < |min|.
   - This is only meaningful if we can guarantee that max >= 0.
   - Rather than relying on finite precision, we change our initializaiton values.
4. Removal of a frivolous test where the candidate points for creating the tet were both on the triangle. I've removed the frivolous test and added documentation for why we don't treat this case as special.

- - -
This change is [Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/658)


sherm1 (approved):
:lgtm: with a few ignorable suggestions
What's the CI story here -- still just your local build?

> @sherm1 reviewed 2 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/658#-:-Ojb_AYWA3p5RrAPPK42:b-p9wdar)* status: all files reviewed, 2 unresolved discussions (waiting on @SeanCurtis-TRI).

___
*[`include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h` line 855 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/658#-OjbXsIG4p889exgwSLi:-OjbXsIG4p889exgwSLj:b-cwfb2w) ([raw file](https://github.com/flexible-collision-library/fcl/blob/95e8b6e8cfbe6a5624e08ff5d8a243d2431b9a22/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L855)):*
> ```objc
> 
>   // The scaled distance of the second candidate,
>   ccdVec3Scale(&dir, -CCD_ONE);
> ```

BTW below (line 1115) you wrote -1 differently. Probably better for them to match? Or maybe you're trying to match the existing inconsistencies?



_Suggestion:_
```objc
ccd_real_t(-1)
```
___
*[`include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h` line 1147 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/658#-OjbYztT1Q-QgnTWcvtt:-OjbYztT1Q-QgnTWcvtu:b-uq54ro) ([raw file](https://github.com/flexible-collision-library/fcl/blob/95e8b6e8cfbe6a5624e08ff5d8a243d2431b9a22/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L1147)):*
> ```objc
>       // (v->v.v - face_point). Note that origin_distance_to_plane = nᵀ *
>       // face_point.
>       const ccd_real_t distance_to_plane =
> ```

BTW I found it a little confusing to follow this code where "distance" means signed offset, especially when maxing and min-ing it below. Writing "signed_distance" everywhere would be awkward, but how about using the term "offset" or something like that instead of distance? Every time I read "distance" I think "positive quantity" and then I have to re-think to follow the code.

SeanCurtis-TRI (commented):
We're not wholly dependent on my machine and my pinky swear. The [windows build](https://ci.appveyor.com/project/flexible-collision-library/fcl/builds/53417203) shows success.

As for the most recent version, I've got local success and we'll defer merging until we get a clean bill of health on windows.

> @SeanCurtis-TRI made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/658#-:-Ojc59FP0VNkijd-A1QK:b-nstwcn)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @sherm1).

___
*[`include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h` line 855 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/658#-OjbXsIG4p889exgwSLi:-Ojc51qBDoCrU6eerI8P:bz2iycg) ([raw file](https://github.com/flexible-collision-library/fcl/blob/95e8b6e8cfbe6a5624e08ff5d8a243d2431b9a22/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L855)):*

BTW below (line 1115) you wrote -1 differently. Probably better for them to match? Or maybe you're trying to match the existing inconsistencies?
</blockquote></details>

The use of CCD_ONE is more typical of the file, so I pushed it that way for consistency.
___
*[`include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h` line 1147 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/658#-OjbYztT1Q-QgnTWcvtt:-Ojc4iNiEvLSUPjlHoFe:b71slvf) ([raw file](https://github.com/flexible-collision-library/fcl/blob/95e8b6e8cfbe6a5624e08ff5d8a243d2431b9a22/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L1147)):*

BTW I found it a little confusing to follow this code where "distance" means signed offset, especially when maxing and min-ing it below. Writing "signed_distance" everywhere would be awkward, but how about using the term "offset" or something like that instead of distance? Every time I read "distance" I think "positive quantity" and then I have to re-think to follow the code.
</blockquote></details>

Actually, "signed distance" wasn't that obnoxious after all. And requires no explanations.

sherm1 (commented):
> @sherm1 reviewed all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/658)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @SeanCurtis-TRI).

___
*[`include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h` line 1147 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/658#-OjbYztT1Q-QgnTWcvtt:-Ojc9YfI8XqTn7UlvzXD:b-wlf23k) ([raw file](https://github.com/flexible-collision-library/fcl/blob/95e8b6e8cfbe6a5624e08ff5d8a243d2431b9a22/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L1147)):*

Actually, "signed distance" wasn't that obnoxious after all. And requires no explanations.
</blockquote></details>

That's better, but there are still `{origin,max,min}_distance_to_plane`. Are those all nonnegative distances? Isn't min_distance_to_plane at least still a signed distance?

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/658)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @sherm1).

___
*[`include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h` line 1147 at r1](https://reviewable.io/reviews/flexible-collision-library/fcl/658#-OjbYztT1Q-QgnTWcvtt:-OjuvpqGEeOuFyb_rrtO:bpx3y0n) ([raw file](https://github.com/flexible-collision-library/fcl/blob/95e8b6e8cfbe6a5624e08ff5d8a243d2431b9a22/include/fcl/narrowphase/detail/convexity_based_algorithm/gjk_libccd-inl.h#L1147)):*

That's better, but there are still `{origin,max,min}_distance_to_plane`. Are those all nonnegative distances? Isn't min_distance_to_plane at least still a signed distance?
</blockquote></details>

I've gone back the other way. Because all evaluated distances in this function are signed distances, I've put a general note in the documentation (in addition to the pre-existing notes calling out signed distance).

sherm1 (approved):
All good -- ready to merge IMO

> @sherm1 reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/658#-:-Ojv6h9Y61KXjazcgPzp:b68ua8s)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SeanCurtis-TRI).

SeanCurtis-TRI (commented):
I think the windows CI is still another 30 minutes away from completion. I'm going to play it safe with such a brittle CI ecosystem.

> @SeanCurtis-TRI made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/flexible-collision-library/fcl/658#-:-OjvBmodEmosg23LSOX1:bb8ih69)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SeanCurtis-TRI).

